PR #3, opened 2021-06-11
Ivan's review on ONEWallet contract

polymorpher:


code/contracts/ONEWallet.sol
```diff
@@ -7,14 +7,15 @@ contract ONEWallet {
...
     uint32 lifespan;  // in number of block (e.g. 1 block per [interval] seconds)
-    uint8 maxOperationsPerInterval; // number of transactions permitted per OTP interval. Each transaction shall have a unique nonce. The nonce is auto-incremented within each interval
+    uint8 maxOperationsPerInterval; // number of transactions permitted per OTP interval. Each transaction shall have a unique nonce. The nonce is auto-incremented within each interval 
 
     // global mutable
     address payable lastResortAddress; // where money will be sent during a recovery process (or when the wallet is beyond its lifespan)
     uint256 dailyLimit; // uint128 is sufficient, but uint256 is more efficient since EVM works with 32-byte words.
     uint256 spentToday; // note: instead of tracking the money spent for the last 24h, we are simply tracking money spent per 24h block based on UTC time. It is good enough for now, but we may want to change this later.
-    uint32 lastTransferDay;
+    uint32 lastTransferDay; // I'd use uint instead
```
polymorpher, line 16: uint32 is enough to represent days. For gas efficiency, I noticed a trend in newer versions of solidity that the gas discrepancy between 256-bit variables and smaller ones are decreasing over time. Some of these can also be packed into a struct to further optimize gas. https://ethereum.stackexchange.com/questions/3067/why-does-uint8-cost-more-gas-than-uint256?noredirect=1&lq=1

code/contracts/ONEWallet.sol
```diff
@@ -7,14 +7,15 @@ contract ONEWallet {
...
 
+    // IH: you can drop nonce stuff if you implement commits as mapping (hash => Commit). After reveal you might delete the Commit entry; the timestamp's fresheness will ensure protection against replay in the future.
```
polymorpher, line 18: It seems still insufficient. Consider the scenario which a commit is revealed during the same 30-second interval, but may span across multiple blocks (recall Harmony generates a block per 2 seconds). If a commit is deleted after the corresponding reveal is completed and there is no nonce tracking, an attacker may immediately recommit another transaction using a correct proof but with modified parameters (e.g. different transfer destination and amount), then immediately reveal it. Since the commit is already deleted and nonce is not used, the smart contract would have no way to identify that this new transaction should be rejected.

code/contracts/ONEWallet.sol
```diff
@@ -24,12 +25,12 @@ contract ONEWallet {
...
     }
 
-    uint32 constant REVEAL_MAX_DELAY = 60;
+    uint32 constant REVEAL_MAX_DELAY = 60; 
 
     //    bool commitLocked; // not necessary at this time
     Commit[] commits; // self-clean on commit (auto delete commits that are beyond REVEAL_MAX_DELAY), so it's bounded by the number of commits an attacker can spam within REVEAL_MAX_DELAY time in the worst case, which is not too bad.
 
-    uint256[64] ______gap; // reserved space for future upgrade
+    uint256[64] ______gap; // reserved space for future upgrade // IH: I am not sure how this work
```
polymorpher, line 33: This is following the practice in https://docs.openzeppelin.com/upgrades-plugins/1.x/writing-upgradeable and https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/v3.3.0/contracts/token/ERC20/ERC20Upgradeable.sol . Unimportant right now, since it is unlikely the wallet instance will be used in another smart contract any time soon. I added this field just to keep this use case in mind.

code/contracts/ONEWallet.sol
```diff
@@ -38,16 +39,16 @@ contract ONEWallet {
         height = height_;
         interval = interval_;
         t0 = t0_;
-        lifespan_ = lifespan_;
+        lifespan_ = lifespan_; // IH: what are the expected values?
```
polymorpher, line 42: It would be (validity time in seconds / interval) where interval should be 30 for standard uses. I intend to use a default value of 1 year for validity time in the client.

code/contracts/ONEWallet.sol
```diff
@@ -38,16 +39,16 @@ contract ONEWallet {
...
         lastResortAddress = lastResortAddress_;
         dailyLimit = dailyLimit_;
-        maxOperationsPerInterval = maxOperationsPerInterval_;
+        maxOperationsPerInterval = maxOperationsPerInterval_; // IH: what are the expected values?
```
polymorpher, line 45: I intend to use 1 for the initial release (to 10 users), and increase to 4 or 16 in subsequent releases.

code/contracts/ONEWallet.sol
```diff
@@ -38,16 +39,16 @@ contract ONEWallet {
...
     }
 
     function retire() external returns (bool)
     {
         require(uint32(block.timestamp) / interval - t0 > lifespan, "Too early to retire");
-        require(lastResortAddress != address(0), "Last resort address is not set");
+        require(lastResortAddress != address(0), "Last resort address is not set"); // IH: it needs method for adjustment; currently, if it adjust it in constructor to 0x0, retire() will never work
```
polymorpher, line 51: Yes, I will add a function that allows setting  lastResortAddress iff it is 0

code/contracts/ONEWallet.sol
```diff
@@ -64,30 +65,33 @@ contract ONEWallet {
         (uint32 ct, bool completed) = _findCommit(hash);
         require(ct == 0 && !completed, "Commit already exists");
         Commit memory nc = Commit(hash, uint32(block.timestamp), false);
-        commits.push(nc);
+        commits.push(nc); // IH: would not it be better to have mapping of commits?
```
polymorpher, line 68: We discussed on Telegram. To include the discussion:

```
I had an early version of commit with map but quick scratched it. The issue is an attacker can commit a lot of garbage (at the expense of their own gas) and make the contract storage unnecessarily large, and unbounded. Thus, there needs to be some mechanisms for removing the map keys. Since we can’t iterate a map in solidity, we have to use an array to track it. Therefore, we have Commit[]. Nice catch with uint8 - forgot to fix that
```

You replied

```
For this cases is usefull EnumerableMap from openzeppelin library. Maybe you can check it
```

I checked EnumerableMap. It seems they rely on using an array to track the keys all the same. The contracts are quite heavy: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/structs/EnumerableSet.sol and https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/structs/EnumerableMap.sol

So I think using a simple array to track and clean the commits are more gas-efficient than using EnumerableMap. Besides, in C++ and Java, looking up an element using linear search over a small array is actually much faster than using a map structure. I don't know if the same would apply to Solidity and EVM, but I suspect it is likely to behave the same.

code/contracts/ONEWallet.sol
```diff
@@ -64,30 +65,33 @@ contract ONEWallet {
...
     }
 
     // TODO (@polymorpher): during reveal, reject if index corresponds to a timestamp that is same as current block.timestamp, so we don't let the client accidentally leak EOTP that is still valid at the current time, which an attacker could use to commit and reveal a new transaction. This introduces a limitation that there would be a OTP_INTERVAL (=30 seconds by default) delay between commit and reveal, which translates to an unpleasant user experience. To fix this, we may slice OTP_INTERVAL further and use multiple EOTP per OTP_INTERVAL, and index the operations within an interval by an operation id. For a slice of size 16, we will need to generate a Merkle Tree 16 times as large. In this case, we would reject transactions only if both (1) the index corresponds to a timestamp that is same as current block.timestamp, and, (2) the operation id is not less than the id corresponding to the current timestamp's slice
+    // IH: what do you state above just supports HOTPs instead of TOTPs, besides other advantages.
```
polymorpher, line 72: We also discussed this in the chat. As I mentioned

> By the way this TODO was already done. Forgot to remove the comment.

Nonetheless, I agree switching to HOTP would eliminate this problem, since we would allow each OTP (and associated proof) to be used exactly once. But the nonce-mechanism implemented here also achieves the same result. My main concern for using a HOTP-based mechanism in the single-factor setting is the following: 

(1) since we have to allow synchronization of counters between smart contract, authenticator, and client, we would also have to always accept HOTPs (and their proofs) and later indices. This means the permission to operate the smart contract wallet is not evenly divided into time-slices, as we would have nicely in the current TOTP-based mechanism. Instead, each index would essentially give a permission to operate the wallet once, but once a later index is used, it would invalidate an earlier index, rendering the earlier index useless.

(2) The client needs to load all hashed, salted HOTPs in proximity to the expected counter, to effectively synchronize the counter with the authenticator. In TOTP based mechanism, the client only needs to load one or two by pinpointing the current time. The difference is not significant at present, but consider in the future we may store the hashed-salted OTPs in a more secure storage, the difference would mean that the former approach (HOTP) would be more vulnerable at the client.

(3) HOTP is much less commonly used. People are much less used to operate on HOTP and the chance of human error is not low. On Google Authenticator, once the current HOTP is displayed, it stays displayed for a long time unless the user switches out from the app, or manually click the refresh button. This would be problematic if the user is using the wallet on the computer, and using the authenticator on the phone. It is likely that the user cannot remember whether the current displayed code is used or not (keep in mind the user would have many other OTPs displayed on the same UI). The user would type in a code that is already used, only to find out it is not accepted, then the user would hit the refresh button and try again. This introduces non-trivial friction and annoyance which should be avoided.

code/contracts/ONEWallet.sol
```diff
@@ -64,30 +65,33 @@ contract ONEWallet {
...
+    // IH: combining index with nonce is unnecessary complicated to me. I would split them.
```
polymorpher, line 73: Together, they represent one position (index) in the leaves. Also, it would save some bytes, since they are implicitly packed together in a uint32.

code/contracts/ONEWallet.sol
```diff
@@ -64,30 +65,33 @@ contract ONEWallet {
...
     function revealTransfer(bytes32[] calldata neighbors, uint32 indexWithNonce, bytes32 eotp, address payable dest, uint256 amount) external
     isCorrectProof(neighbors, indexWithNonce, eotp)
     returns (bool)
     {
-        bytes memory packedNeighbors = _pack(neighbors);
+        bytes memory packedNeighbors = _pack(neighbors); // IH: no need to put Merkle proof into the hash (it is either correct and matches to OTP or not, hence integrity is implicit)
         bytes memory packed = bytes.concat(packedNeighbors,
             bytes32(bytes4(indexWithNonce)), eotp, bytes32(bytes20(address(dest))), bytes32(amount));
-        bytes32 commitHash = keccak256(bytes.concat(packed));
+        bytes32 commitHash = keccak256(bytes.concat(packed)); // IH: why do you need bytes.concat again?
```
polymorpher, line 81: `bytes.concat` concatenates the bytes of each parameter together, without any padding. So we can check the parameters in the reveal call is same as what's committed before. `bytes.concat` is chosen over other functions because it is simple and perhaps is the easiest to replicate in javascript, whereas alternatives such as abi.encodePacked introduces padding and many other complicated behaviors which are not necessary for our use case.

code/contracts/ONEWallet.sol
```diff
@@ -64,30 +65,33 @@ contract ONEWallet {
...
         _revealPreCheck(commitHash, indexWithNonce);
         _completeReveal(commitHash);
-        uint32 day = uint32(block.timestamp) / 86400;
+        uint32 day = uint32(block.timestamp) / 86400; // IH: goof to use a constant; also casting uint 256 as uint32 does not seem reasonable (maybe if you'd put division into the parenthesis). But anyway, I'd go for uint instead.
```
polymorpher, line 84: Yes, I will move the constant outside and put it in a constant variable.  As for block.timestamp casting into uint32 - it should be good until year ~2100 (max uint value = 4.2e+9, current timestamp in seconds: 1.6e+9). Moving 86400 inside should ensure absolute safety. Also I think uint32 would be more efficient than uint256 (uint) here since this is in-memory rather than using gas in storage.

code/contracts/ONEWallet.sol
```diff
@@ -64,30 +65,33 @@ contract ONEWallet {
...
         if (day > lastTransferDay) {
             spentToday = 0;
             lastTransferDay = day;
         }
-        if (spentToday + amount > dailyLimit) {
+        if (spentToday + amount > dailyLimit) { // IH: I'd put revert here. So the user will be warned by client that tx will not pass and do not proceed => save gas.
```
polymorpher, line 89: We should not revert here, because it would undo the reveal (and nonce increment), making it possible to re-use the same OTP and its proof even though it is already revealed exposed.

code/contracts/ONEWallet.sol
```diff
@@ -64,30 +65,33 @@ contract ONEWallet {
...
             return false;
         }
-        bool success = dest.send(amount);
-        // we do not want to revert the whole transaction if this operation fails, since EOTP is already revealed
+        // IH: you should check enough balance first
```
polymorpher, line 92: yes, though `.send` should fail implicitly and return false if there is not enough balance

code/contracts/ONEWallet.sol
```diff
@@ -105,13 +109,15 @@ contract ONEWallet {
             bytes32(bytes4(indexWithNonce)),
             eotp
         );
-        bytes32 commitHash = keccak256(bytes.concat(packed));
+        bytes32 commitHash = keccak256(bytes.concat(packed)); // IH: why do you need bytes.concat again?
```
polymorpher, line 112: Same as above

code/contracts/ONEWallet.sol
```diff
@@ -105,13 +109,15 @@ contract ONEWallet {
...
         _revealPreCheck(commitHash, indexWithNonce);
         _completeReveal(commitHash);
+        // IH: until here, the code is duplicate with revealTransfer(). Can you put it into some helper function?
```
polymorpher, line 115: There isn't a way to get a list of arguments, otherwise we could use a common modifier

code/contracts/ONEWallet.sol
```diff
@@ -105,13 +109,15 @@ contract ONEWallet {
...
+
         if (lastResortAddress == address(0)) {
             return false;
         }
-        _drain();
+        _drain(); // IH: the function is short, can you expand it here? and also check return code?
```
polymorpher, line 120: Yes, it should return `_drain()` rather than ignoring the return value

code/contracts/ONEWallet.sol
```diff
@@ -145,12 +151,13 @@ contract ONEWallet {
         _;
     }
 
+    // IH: why to not make just mapping of hashes to Commits? Then you might drop this and the next functions. Further, you won't have out-of-gas DOS by the attacker in commit stage => manifested here.
```
polymorpher, line 154: Discussed above

code/contracts/ONEWallet.sol
```diff
@@ -145,12 +151,13 @@ contract ONEWallet {
...
     function _findCommit(bytes32 hash) view internal returns (uint32, bool)
     {
         if (hash == "") {
             return (0, false);
         }
-        for (uint8 i = 0; i < commits.length; i++) {
+        for (uint8 i = 0; i < commits.length; i++) { // IH: uint8 might not be enough, I'd just go for uint since it is (cheap) in-memory var.
```
polymorpher, line 160: Yes, should be uint32. The point on optimization is discussed above. Not sure whether in-memory operations are optimized at this point for shorter ints over longer ones (uint256), if not, I think Solidity should really fix it in future versions.

code/contracts/ONEWallet.sol
```diff
@@ -206,12 +214,12 @@ contract ONEWallet {
         (uint32 ct, bool completed) = _findCommit(hash);
         require(ct > 0, "Cannot find commit for this transaction");
         uint32 counter = ct / interval - t0;
-        require(counter == index, "Provided index does not match commit timestamp");
+        require(counter == index, "Provided index does not match committed timestamp");
         uint8 expectedNonce = nonces[counter];
-        require(nonce >= expectedNonce, "Nonce is too low");
+        require(nonce >= expectedNonce, "Nonce is too low"); // IH: should not be here equality? One commit object is associated with the one nonce (it is part of commited hash)
```
polymorpher, line 219: Should be `>=` because nonce is incremented during reveal after passing `_revealPreCheck`. Nonce helps tracking the number of proofs already used within one interval. It starts with 0.

code/contracts/ONEWallet.sol
```diff
@@ -206,12 +214,12 @@ contract ONEWallet {
...
         require(!completed, "Commit is already completed");
         // this should not happen (since old commit should be cleaned up already)
-        require(_isRevealTimely(ct), "Reveal is too late. Please re-commit");
+        require(_isRevealTimely(ct), "Reveal is too late. Please re-commit"); // IH: What might happen if we'd increase REVEAL_MAX_DELAY 10x or 100x? I do not see any advantage of keeping it short since MITM attacker gets only extended 32B OTP while attacker that tampers with the client can do anything regardless of this lenght. This brings me again to using HOTPs.
```
polymorpher, line 222: That would make the storage for commit very large when an attacker commits a lot of garbage. It is not much of a problem if it is kept temporary, since the attacker paid for it. But it must not be made permanent (or stays too long, in the case of making REVEAL_MAX_DELAY 10x or 100x), otherwise the size of the contract grows unbounded without producing additional value. In practice, the reveal should happen immediately after a commit. It is not reasonable for a client to wait for too long after committing an operation aside from bad network conditions.

code/contracts/ONEWallet.sol
```diff
@@ -263,7 +271,7 @@ contract ONEWallet {
         nonces[index] = v + 1;
     }
 
-    // same output as abi.encodePacked(x), but the behavior of abi.encodePacked is sort of uncertain
+    // same output as abi.encodePacked(x), but the behavior of abi.encodePacked is sort of uncertain - IH: this is surprising; in which sense?
```
polymorpher, line 274: Based on the spec here https://docs.soliditylang.org/en/latest/abi-spec.html#non-standard-packed-mode , arrays may get padded with `abi.encodePacked`.  This is not the case here, but I do not want to rely on this function since its behavior may change in the future. Our use case is pretty simple: given a list of parameters, convert them into 32-byte elements and concatenate them together. The whole `abi` encoding spec seems to be an overkill for this use case.

code/contracts/ONEWallet.sol
```diff
@@ -64,30 +65,33 @@ contract ONEWallet {
...
+    // IH: what do you state above just supports HOTPs instead of TOTPs, besides other advantages.
+    // IH: combining index with nonce is unnecessary complicated to me. I would split them. 
     function revealTransfer(bytes32[] calldata neighbors, uint32 indexWithNonce, bytes32 eotp, address payable dest, uint256 amount) external
     isCorrectProof(neighbors, indexWithNonce, eotp)
     returns (bool)
     {
-        bytes memory packedNeighbors = _pack(neighbors);
+        bytes memory packedNeighbors = _pack(neighbors); // IH: no need to put Merkle proof into the hash (it is either correct and matches to OTP or not, hence integrity is implicit)
```
polymorpher, line 78: That's true. However, if we remove the `neighbors` from the packed bytes computation, the only issue seems to be that an attacker could commit the exact same transaction ahead of the user, resulting the user receiving an annoying rejection message "commit already exists" when the user truly commits it. This may result in a bad user experience. Since only neighbors[0] is secret, we could keep that inside the `packed` instead of the whole `neighbors`. This way the attacker would have no way of committing the same transaction ahead of the user since they would not know the value of `neighbors[0]`

code/contracts/ONEWallet.sol
ivan-homoliak-sutd: ok, stay with uint8, just be sure that it will not overflow.

code/contracts/ONEWallet.sol
ivan-homoliak-sutd: Yes, you are right. I've somehow already assumed HOTPs, in which this cannot not happen by default since SC remembers ids of the used ones (or id of the last used one) while allowing only not yet used (or later ones). But going back to TOTPs - I'd still suggest to drop the nonce stuff (as too verbose) and do the things simpler. So, to resolve the issue with deleting from the map of commits, I'd use a dedicated cleanup external method (called once in a while), while using another EnumerableMap (EOTPs => timestamps) to track EOTPs to delete, and their timestamps of expiration. In this method, SC would just iterate over EOTPs in the map and check the time validity of EOTP according to block.timestamp, followed by the delete from both maps with the purpose of saving the storage space. Also, in any revealX(), SC would need in addition to insert data into the second map.

code/contracts/ONEWallet.sol
polymorpher: fixed in https://github.com/polymorpher/one-wallet/commit/d289e3e3802c664ab73d97ec5aa79631f896af0f

code/contracts/ONEWallet.sol
polymorpher: fixed in https://github.com/polymorpher/one-wallet/commit/d289e3e3802c664ab73d97ec5aa79631f896af0f

code/contracts/ONEWallet.sol
polymorpher: fixed in https://github.com/polymorpher/one-wallet/commit/d289e3e3802c664ab73d97ec5aa79631f896af0f

code/contracts/ONEWallet.sol
polymorpher: addressed in https://github.com/polymorpher/one-wallet/commit/d289e3e3802c664ab73d97ec5aa79631f896af0f

code/contracts/ONEWallet.sol
polymorpher: addressed in https://github.com/polymorpher/one-wallet/commit/d289e3e3802c664ab73d97ec5aa79631f896af0f

code/contracts/ONEWallet.sol
```diff
@@ -24,12 +25,12 @@ contract ONEWallet {
...
-    uint32 constant REVEAL_MAX_DELAY = 60;
+    uint32 constant REVEAL_MAX_DELAY = 60; 
 
     //    bool commitLocked; // not necessary at this time
     Commit[] commits; // self-clean on commit (auto delete commits that are beyond REVEAL_MAX_DELAY), so it's bounded by the number of commits an attacker can spam within REVEAL_MAX_DELAY time in the worst case, which is not too bad.
 
-    uint256[64] ______gap; // reserved space for future upgrade
+    uint256[64] ______gap; // reserved space for future upgrade // IH: I am not sure how this work
 
     constructor(bytes32 root_, uint8 height_, uint8 interval_, uint32 t0_, uint32 lifespan_, uint8 maxOperationsPerInterval_,
```
polymorpher, line 35: The contract was not payable. This is addressed in https://github.com/polymorpher/one-wallet/commit/4f3f9df87b5f7570feb4af68e461390a0d4a2e64

code/contracts/ONEWallet.sol
polymorpher: This is feasible. One issue is the caller of the external method would have to pay for the gas to cleanup. The attacker should pay for that, if an attack exists. But this can be mitigated by making this method internal and call it sporadically as part of the commit process.

But the primary purpose of nonce is to enable multiple operations per interval, not tracking or managing commits. Each nonce value corresponds to an EOTP inside an interval. Each interval has one (T)OTP, but has maxOperationsPerInterval number of EOTPs (corresponding to each nonce). The current method does not track used EOTPs. It simply requires a transaction to use a higher nonce than before (thus must correspond with different EOTP).

The method for managing and tracking commits are independent from nonce.

hashmesan (commented):
I see the introduction of indexWithNonce. Maybe some comment and justification for this would be helpful. I don't understand it yet.   And transfer will need to support other ERC-20 tokens in the future as well.

code/contracts/ONEWallet.sol
ivan-homoliak-sutd: It's at the attacker's expense anyway. Therefore, the attacker is not incentivized to do it since he pays for it. And Harmony blockchain wants users who pays for txs. Hence, to me, growing blockchain is an unrelated issue - such attacker will not occur for sure (unless something is improperly designed - however, this we can influence).

code/contracts/ONEWallet.sol
ivan-homoliak-sutd: I am sure that its behavior will not change. If anything like that would be about to happen, not in the same version of solidity, so it would not impact us. This function provides unambiguous concatenation and I've seen it in many other contracts.

code/contracts/ONEWallet.sol
ivan-homoliak-sutd: OK

code/contracts/ONEWallet.sol
polymorpher: Marking this as resolved

code/contracts/ONEWallet.sol
polymorpher: Will remove soon, since we shouldn't need to worry about upgrade using contract at this time. We can do it later when users report this as an issue

code/contracts/ONEWallet.sol
polymorpher: One year is more than enough for the basic wallet. The feedback I got is to make it even shorter, since we are going to upgrade the wallet pretty soon anyway. Later we can provide a way to refresh wallet lifespan after the user activates another authentication factor.

code/contracts/ONEWallet.sol
polymorpher: Okay, I will add a lookup map. I want to minimize the use of external dependencies at the core level, unless we have complex logics that really require these libraries. They make debugging and analysis harder, and uses more gas.

code/contracts/ONEWallet.sol
polymorpher: On (1), keep in mind we are still limited by the (Google) authenticator, which does not take variable input when it is generating code. The user may randomly refresh (hence increment counters) on the authenticator, hence introducing the out-of-sync problem, and force us to allow skipping to later EOTP (thus attacker only need to crack a later one, rather than being limited by a time window)

On (2) - you would be correct if we consider client as a whole - i.e. either all information at the client is exposed, or nothing is exposed. I think there should exist mechanisms which we can store majority of the information (e.g. leaves not relevant to current time) in a secure storage (e.g. a vault or enclave), and only load as minimally as possible for whatever is needed as the user executes an operation. The mechanism is still theoretical in browser, but it is quite practical if we venture into the realm of native applications (on either desktop or mobile).

code/contracts/ONEWallet.sol
polymorpher: You are right. I somehow thought EOTP would be the same for two different nonces. They are not. So EOTP alone is sufficient. But based on our discussion (outside this PR) we are going to rewrite the logic for commit-reveal anyway to cure one of the attacks.

code/contracts/ONEWallet.sol
polymorpher: Just did a test - concat costs slightly less gas (but the difference is really small). I am in favor of concat because it is more predictable and is much less complex. On Javascript side the bytes manipulation is also much simpler and cleaner than using external dependencies (web3 abi library)

code/contracts/ONEWallet.sol
polymorpher: Let's stick to simpler expression on both contract and Javascript sides for now. The Javascript side of the code for abi.encodePacked looks quite messy.

code/contracts/ONEWallet.sol
polymorpher: On a second thought, perhaps we should allow an "untimely" reveal. There are legitimate use cases for this in dApps: sign a transaction at first, execute it later. The smart contract wallet equivalent would be to commit (with signature) at first, reveal later.

polymorpher:
Okay. I will update the specs soon to properly document new concepts. Let's wrap up this review and move on to newer versions, since many things will change.

polymorpher:
Let's reopen this when Ivan is back from holiday

PR #4, opened 2021-06-23
An example of a shortened ONE Wallet contract with equivalent functionalities

ivan-homoliak-sutd:


code/contracts/ONEWallet2.sol
```diff
@@ -0,0 +1,157 @@
...
+            } else {
+                h = sha256(abi.encodePacked(h, neighbors[i]));
+            }
+            index >>= 1;
+        }
+        require(root == h, "Proof is incorrect");        
+    }
+
+    function _executeTransfer(address payable dest, uint amount) internal {
+        require(dailySpendings[block.timestamp / 1 days] + amount <= dailyLimit, "Daily limit exceeded.");
```
polymorpher, line 149: This and below statement would revert the reveal. The attacker can use the leaked proofs to commit and reveal his altered transaction.

code/contracts/ONEWallet2.sol
```diff
@@ -0,0 +1,157 @@
...
+    uint32 t0; // starting time block (effectiveTime (in ms) / interval)
+    uint32 lifespan;  // in number of blocks (e.g. 1 block per [interval] seconds)
+    
+    address payable lastResortAddress; // where money will be sent during a recovery process (or when the wallet is beyond its lifespan)
+    bytes32 recoveryHash; // hash of the secret stored at the client, which can be used to recover the walletif the user loses the authenticator
+    uint8 constant MAX_REVEAL_DELAY = 2; // in OTP intervals (i.e., 2 * interval seconds)    
+    
+    // Management of daily spendings
+    uint256 dailyLimit; // uint128 is sufficient, but uint256 is more efficient since EVM works with 32-byte words.
+    mapping(uint => uint) dailySpendings; // day => amount ; can be cleaned up once in a while to save storage
```
polymorpher, line 18: We shouldn't  need to track this by day since only the latest matter and the information can be reconstructed on-demand by querying past transactions of the contract address anyway.

code/contracts/ONEWallet2.sol
```diff
@@ -0,0 +1,157 @@
...
+    uint cleanupCheckpoint; // the index of the day since epoch, when the last cleanup was made
+    uint16 constant CLEAN_UP_DAILY_SPENDING_BATCH = 1024; // this is to avoid out-of-gas exceptions
+    uint8 AUTO_CLEANUP_DAYS = 100; // do the auto cleanup of daily spending data after every 100 days
+    
+    uint lastExecutedIdx; // stores the index of OTP that was executed last (this is to avoid reply attacks within the same OTP interval by not allowing commits <= than this number)
+
+    enum OperType { TRANSFER, RECOVERY, SET_LAST_RESORT_ADDR }
+
+    struct Commit {
```
polymorpher, line 27: I think we should hide the nature of the commit as much as possible. A user's intention should not be revealed in many use cases. For example, in a bidding or betting scenario, the time gap between an exposed commit and a reveal would put the user at some disadvantage.

code/contracts/ONEWallet2.sol
```diff
@@ -0,0 +1,157 @@
...
+        bytes32 hash;          // h(eOTP || operType || params)  
+        uint timestamp;      // OTP must be valid in this time; not in the time of reveal        
+        OperType operType;
+        // below are operation-specific parameters, based on operType
+        address payable addr; 
+        uint amount;        
+    }    
+    mapping(uint32 => Commit[]) public commits; // idx of OTP valid during a commitment => commitment data
```
polymorpher, line 35: I noticed only revealed commits are deleted (line 103). The commits would grow unbounded if an attacker spams it.

code/contracts/ONEWallet2.sol
```diff
@@ -0,0 +1,157 @@
...
+        uint32 submittedOTPIdx = uint32(block.timestamp) / interval - t0;
+        require(submittedOTPIdx > lastExecutedIdx, "It is not possible to submit commits in the current OTP interval anymore. Wait for the next interval.");
+        Commit memory nc = Commit(hash, block.timestamp, otype, addr, amount);
+        commits[submittedOTPIdx].push(nc);                
+    }
+
+    /**
+     * This is a general reveal for all types of operation.
+     */
+    function reveal(bytes32[] calldata neighbors, uint32 otpIdx, bytes32 eotp) external  returns (bool) {
```
polymorpher, line 68: I think this is a nice way of unifying the threes reveal functions into a single one. As mentioned above if we want to hide the nature of the commit, we would then need to add another parameter here (a struct that unifies parameters of all kinds of operations)

code/contracts/ONEWallet2.sol
```diff
@@ -0,0 +1,157 @@
...
+    uint16 constant CLEAN_UP_DAILY_SPENDING_BATCH = 1024; // this is to avoid out-of-gas exceptions
+    uint8 AUTO_CLEANUP_DAYS = 100; // do the auto cleanup of daily spending data after every 100 days
+    
+    uint lastExecutedIdx; // stores the index of OTP that was executed last (this is to avoid reply attacks within the same OTP interval by not allowing commits <= than this number)
+
+    enum OperType { TRANSFER, RECOVERY, SET_LAST_RESORT_ADDR }
+
+    struct Commit {
+        bytes32 hash;          // h(eOTP || operType || params)  
+        uint timestamp;      // OTP must be valid in this time; not in the time of reveal
```
polymorpher, line 29: By the way timestamp (uint32), operationType (byte), and address (bytes20) can be packed into one bytes32. It should reduce storage cost by 40% under this scheme.

code/contracts/ONEWallet2.sol
```diff
@@ -0,0 +1,157 @@
...
+                } else if(c.operType == OperType.SET_LAST_RESORT_ADDR){
+                    if(lastResortAddress == address(0x0) && c.addr != address(0x0))
+                        lastResortAddress = c.addr;
+                } else if(c.operType == OperType.RECOVERY){ // we do not care about params in Commits
+                    _drain(); // IH: I do not find any reason for this operation to be executed using OTP.
+                } else {
+                    emit UnknownOperationType(c.operType);                    
+                }                
+                
+                delete commits[otpIdx][i];
```
polymorpher, line 103: This and the next line should be above the execution block (line 90). So even if the execution becomes out-of-gas , the commit can be deleted and lastExecutedIdx is properly set to new index.

polymorpher:
I will incorporate some good practices into the contract on master branch

polymorpher:
Let's reopen this when Ivan is back from holiday

code/contracts/ONEWallet2.sol
ivan-homoliak-sutd: I do not agree. All transactions are **atomic**, so they are either fully executed or not executed at all. If there is out-of-gas error, everything is just reverted to its original state.

code/contracts/ONEWallet2.sol
ivan-homoliak-sutd: You are ritght. Good catch! Will fix it soon.

ivan-homoliak-sutd (commented):
Reactions of comments of the simple design. There are two issues to me: 1) reverting in execute transfer needs to be removed and added return statement instead. 2) if the commit really needs to hide the nature of the operation (type, value etc), the design needs to be modified. However, so far I do not see any use case that would require it.

code/contracts/ONEWallet2.sol
polymorpher: Resolved, since this seems to have been updated. Previously, with the parameters exposed, the issue is that an adversary knowing a user's intent (and exact amount) to bid, could carefully place a slightly higher bid between commit and reveal. A similar scenario can be formulated in adversarial games. The user who (is forced to) announces its intent earlier than others are put at a disadvantage.

code/contracts/ONEWallet2.sol
polymorpher: It clutters the blockchain. But another potential danger is if we are to use an upgrade pattern in this contract. The attacker who is allowed to make unlimited commits may find a way to override key storage locations through spamming commits, though the exact formulation of the attack is unknown at this time due to the assumed reliability of pseudo-randomness of SHA3.

code/contracts/ONEWallet2.sol
```diff
@@ -0,0 +1,176 @@
...
+     * This is a general reveal for all types of operation. Arguments contain all possible parameters as the union (only the relevant args for a particular operation are used).
+     * It cannot revert with not yet invalidated OTP, since the attacker can steal it + its  proof and misuse it for the malicious commit() and reveal().
+     */
+    function reveal(bytes32[] calldata neighbors, uint32 otpIdx, bytes32 eotp, OperType operType, address payable addr, uint amount) external returns (bool) {                        
+
+        // 1) Check time constrains of reveal
+        uint32 currentIdx = uint32(block.timestamp) / interval - t0; // check whether idx of OTP is timely
+        require(currentIdx - otpIdx <= MAX_REVEAL_DELAY, "Reveal is out of max allowed delay."); // here is OK to revert, coz if revert fails, submitted eotp + proof are invalidated already and cannot be misused anymore.
+
+        // 2) scan all the commits made within otpIdx interval and execute the first matching one
```
polymorpher, line 72: Assume a user made a commit and revealed within the same otpIdx interval (30 seconds in general, which is necessary for good user experience). After observing the revealed EOTP but before it is confirmed and executed by the blockchain nodes, an attacker could make another commit with the correct EOTP using different parameters of `operType`, `addr`, and `amount` and produce a valid hash in that commit. The attacker can then reveal its own commit at a higher gas cost before the user's reveal is processed. The attacker's reveal will succeed because it meets the requirement here: under the attacker's parameters, it will be the `first matching commit within the otpIdx interval`.

code/contracts/ONEWallet2.sol
```diff
@@ -0,0 +1,176 @@
...
+    //////////////////////// Public Aux Functions ////////////////////////// 
+
+    /**
+     * It serves for the deletion of accidentaly submitted commits by the user or malicious commits submitted by the attackers, which contribute to the total storage allocation.
+     * Currently, this method can be called by anybody since it does touch only the garbage. 
+     * Also, in theory, someone may earn gas if there are too many entries submitted by the attacker (who payed for it).
+     */
+    function cleanUpCommits() external {                 
+        for (uint32 i = lastCleanUpOfCommitsIdx; i <= lastRevealedOTPIdx; i++){ // IH: in theory, this might lead to out-of-gas exception for settings of TOTP with very long validity but currently I'd not be affraid. Moreover, we are  getting the gas back for successfull deletions.
+            delete commits[i]; // NOTE: I'd say it makes no difference whether it was initialized or not, so we can delete any entry.
```
polymorpher, line 119: For primitive arrays, a deletion on array itself is sufficient to delete everything. But I am uncertain whether it would be sufficient for struct arrays.

code/contracts/ONEWallet2.sol
```diff
@@ -0,0 +1,176 @@
...
+            index >>= 1;
+        }
+        require(root == h, "Proof is incorrect");        
+    }
+
+    function _executeTransfer(address payable dest, uint amount) internal returns(bool) {
+        if(dailySpendings[block.timestamp / 1 days] + amount > dailyLimit)
+            return false; // Daily limit exceeded.        
+        
+        if(!dest.send(amount))
```
polymorpher, line 166: `send` only forward 2300 gas, which is insufficient if `dest` is a contract and implements non-trivial logics in `receive()`. I changed it to `.call{value: amount}()` in my implementation in an early version.

code/contracts/ONEWallet2.sol
```diff
@@ -0,0 +1,176 @@
...
+            return false; // Unknown transfer error.
+                 
+        dailySpendings[block.timestamp / 1 days] += amount;
+        return true;
+    }  
+
+    function _drain() internal returns (bool) {
+        return lastResortAddress.send(address(this).balance);
```
polymorpher, line 174: similar issue with `send` as above

code/contracts/ONEWallet2.sol
```diff
@@ -0,0 +1,176 @@
...
+
+    function retire() external returns (bool){
+        require(uint32(block.timestamp / interval) - t0 > lifespan, "Too early to retire");        
+        return _drain();
+    }
+
+    //////////////////////// Internal Functions ////////////////////////// 
+
+    /**
+     * It cleans up the previous value of daily spending from the map (this ensures that map has always only a single value -- except the execution time of the reveal())
```
polymorpher, line 136: This procedure essentially reduces dailySpendings to a single value, so it needs not to be a mapping anymore

code/contracts/ONEWallet2.sol
polymorpher: Yes, that's true. Maybe I was half asleep when I wrote that :) I forgot the context. Anyway, this issue already becomes moot in the new version.

code/contracts/ONEWallet2.sol
polymorpher: That's true. I am avert to using mapping unless necessary because it stores data at pseudo-random location and requires random access, also costs more gas. But in practice it should not make much difference.

code/contracts/ONEWallet2.sol
ivan-homoliak-sutd: That is true. It was very likely the reason why I put the parameters of an operation into the commit() in the original version instead of reveal(). The first matching one will be always the user's operation. Unfortunately, I did not make any comment to the code about it. So, with parameters in the commit(), it is secure but reveals user intention about the operation, which I still think does cause any problems in the scope of the standard wallet use cases.

code/contracts/ONEWallet2.sol
ivan-homoliak-sutd: The compiler did not complain about it.

code/contracts/ONEWallet2.sol
ivan-homoliak-sutd: I'd say we can safely assume that sha3 is collision-resistant and thus provides enough pseudo-randomness.

code/contracts/ONEWallet2.sol
ivan-homoliak-sutd: The gas cost of using maps is really negligible. That I'd not worry about. I slightly updated it in the recent version, though I still kept the map there. However, it can have only a single element at the time (except for the reveal execution which checks it for the current day and deletes the last day with the transfer). So far, it is still the clearest solution I see.

code/contracts/ONEWallet2.sol
polymorpher: The spec reads:
> `delete a` assigns the initial value for the type to `a`. I.e. for integers it is equivalent to `a = 0`, but it can also be used on arrays, where it assigns a dynamic array of length zero or a static array of the same length with all elements set to their initial value
(https://docs.soliditylang.org/en/v0.8.4/types.html#delete)

It is unclear from the language whether a dynamic array's elements would be set to their initial value

code/contracts/ONEWallet2.sol
polymorpher: Then the current implementation (#v0.4.1, also current master branch) would be similar, except keys of `commits` are encoded as `commitHash` to hide time information and to prevent attacker from committing at future slots, and parameters are also encoded as `paramsHash` to hide user's intent, and to make the structure work for all types of operations in addition to `transfer`.

code/contracts/ONEWallet2.sol
ivan-homoliak-sutd: Proper clean-up method was already added.

code/contracts/ONEWallet2.sol
ivan-homoliak-sutd: IMPORTANT: My conclusion is that we have to store parameters of the operation along with the hash of the commit because the smart contract has to verify in reveal() that h(submitted parameters || eotp) == submitted hash. This is how the first matching entry is found. Actually, this approach is the simplified 3-stage protocol for replacing the root hash in SmartOTPs. While I was designing our commit/reveal, I realized that 3-stages can be reduced to 2-stage, where in 1st stage, you submit hash of commit + params.

This has an implication of showing the parameters of the operation, as mentioned above (which should not be an issue).

code/contracts/ONEWallet2.sol
ivan-homoliak-sutd: I see. Fixed.

polymorpher:
I am merging this into a separate folder - `alternatives` so we can track alternatives considered

PR #15, opened 2021-06-28
Added beta pill to sidenav

givp:
image


polymorpher:
antd also has a tag component which would be helpful to keep the design consistent across the frontend

https://ant.design/components/tag/

givp:
> antd also has a tag component which would be helpful to keep the design consistent across the frontend
> 
> https://ant.design/components/tag/

Ah, wasn't aware. Will replace with the antd component on next PR.

PR #17, opened 2021-06-28
Giv/7 recovery address addition

givp:
@polymorpher I've added a modal for setting the wallet's recovery address (if not already set).

I'm not sure what needs to happen to set the address in the contract from here. I'm happy to update it if you point me in the right direction. I'm also happy for you to take over from here. Whichever you're more comfortable with 👍 

Screen Recording 2021-06-28 at 03 46 25 PM


polymorpher:
Looks good, thanks. I will add a route in relayer and test it at first. With that route you can complete the API call

polymorpher:
Here is the new api for that: https://github.com/polymorpher/one-wallet/commit/7a3be18efb5a6731803d253c8b40307c5139524c

givp:
> Here is the new api for that: [7a3be18](https://github.com/polymorpher/one-wallet/commit/7a3be18efb5a6731803d253c8b40307c5139524c)

Nice, I'll wire it up.

polymorpher:
Well it is a lot more complicated than I anticipated. But I got it done.

polymorpher:
On the UI side I made a few adjustments: instead modal I continued using AnimatedSection in the same style as Transfer screen, so we can keep things consistent. Also it needs an OTP code.

PR #21, opened 2021-06-30
Giv/8 one address

givp:
I've tested with both one and eth addresses and they work interchangeably everywhere.

### DIFF
--- a/code/client/src/pages/Create.jsx
+++ b/code/client/src/pages/Create.jsx
@@ -109,8 +109,8 @@ const Create = () => {
     }
 
     // Ensure valid address for both 0x and one1 formats
-    const validLastResortAddress = util.validateAddress(lastResortAddress)
-    if (!validLastResortAddress) {
+    const normalizedAddress = util.normalizedAddress(lastResortAddress)
+    if (!normalizedAddress) {
       return
     }
 
@@ -123,7 +123,7 @@ const Create = () => {
         t0: effectiveTime / WalletConstants.interval,
         lifespan: duration / WalletConstants.interval,
         slotSize,
-        validLastResortAddress,
+        normalizedAddress,
         dailyLimit: ONEUtil.toFraction(dailyLimit).toString()
       })
       console.log('Deployed. Received contract address', address)
@@ -133,7 +133,7 @@ const Create = () => {
         root: ONEUtil.hexView(root),
         duration,
         effectiveTime,
-        validLastResortAddress,
+        normalizedAddress,
         dailyLimit: ONEUtil.toFraction(dailyLimit).toString(),
         hseed: ONEUtil.hexView(hseed),
         network
--- a/code/client/src/pages/Create.jsx
+++ b/code/client/src/pages/Create.jsx
@@ -123,7 +123,7 @@ const Create = () => {
         t0: effectiveTime / WalletConstants.interval,
         lifespan: duration / WalletConstants.interval,
         slotSize,
-        normalizedAddress,
+        lastResortAddress: normalizedAddress,
         dailyLimit: ONEUtil.toFraction(dailyLimit).toString()
       })
       console.log('Deployed. Received contract address', address)
@@ -133,7 +133,7 @@ const Create = () => {
         root: ONEUtil.hexView(root),
         duration,
         effectiveTime,
-        normalizedAddress,
+        lastResortAddress: normalizedAddress,
         dailyLimit: ONEUtil.toFraction(dailyLimit).toString(),
         hseed: ONEUtil.hexView(hseed),
         network
--- a/code/client/src/pages/Create.jsx
+++ b/code/client/src/pages/Create.jsx
@@ -107,23 +107,13 @@ const Create = () => {
       message.error('Cannot deploy wallet. Error: root is not set.')
       return
     }
-    // Check for one address
-    let verifiedLastResortAddress = lastResortAddress
-    if (verifiedLastResortAddress.startsWith('one')) {
-      if (!HarmonyAddress.isValidBech32(verifiedLastResortAddress)) {
-        console.error(`Invalid address: ${verifiedLastResortAddress}`)
-        return
-      }
-      verifiedLastResortAddress = fromBech32(verifiedLastResortAddress)
-    } else if (verifiedLastResortAddress.startsWith('0x')) {
-      if (!HarmonyAddress.isValidBech32(toBech32(verifiedLastResortAddress))) {
-        console.error(`Invalid address: ${verifiedLastResortAddress}`)
-        return
-      }
-    } else {
-      console.error(`Invalid address: ${verifiedLastResortAddress}`)
+
+    // Ensure valid address for both 0x and one1 formats
+    const validLastResortAddress = util.validateAddress(lastResortAddress)
+    if (!validLastResortAddress) {
       return
     }
+
     setDeploying(true)
     try {
       const { address } = await api.relayer.create({
@@ -133,7 +123,7 @@ const Create = () => {
         t0: effectiveTime / WalletConstants.interval,
         lifespan: duration / WalletConstants.interval,
         slotSize,
-        verifiedLastResortAddress,
+        validLastResortAddress,
         dailyLimit: ONEUtil.toFraction(dailyLimit).toString()
       })
       console.log('Deployed. Received contract address', address)
@@ -143,7 +133,7 @@ const Create = () => {
         root: ONEUtil.hexView(root),
         duration,
         effectiveTime,
-        verifiedLastResortAddress,
+        validLastResortAddress,
         dailyLimit: ONEUtil.toFraction(dailyLimit).toString(),
         hseed: ONEUtil.hexView(hseed),
         network
--- a/code/client/src/pages/Create.jsx
+++ b/code/client/src/pages/Create.jsx
@@ -15,10 +15,10 @@ import storage from '../storage'
 import walletActions from '../state/modules/wallet/actions'
 import WalletConstants from '../constants/wallet'
 import util from '../util'
+import { handleAPIError, handleAddressError } from '../handler'
 import { Hint, Heading, InputBox } from '../components/Text'
-import OtpInput from 'react-otp-input'
 import OtpBox from '../components/OtpBox'
-import { fromBech32, HarmonyAddress, toBech32, getAddress } from '@harmony-js/crypto'
+import { getAddress } from '@harmony-js/crypto'
 const { Text, Link } = Typography
 
 const genName = () => uniqueNamesGenerator({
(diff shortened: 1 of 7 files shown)


code/client/src/pages/Create.jsx
```diff
@@ -106,6 +107,23 @@ const Create = () => {
       message.error('Cannot deploy wallet. Error: root is not set.')
       return
     }
+    // Check for one address
+    let verifiedLastResortAddress = lastResortAddress
```
polymorpher, line 111: Can we makes this a utility in util.js with boolean return value? Also, it should show some feedback to the user (e.g. message.error) at appropriate time (e.g. when full length is reached) when the address is incorrect. It should also give some visual signal when the full length is yet to be reached.

code/client/src/pages/Show.jsx
```diff
@@ -131,11 +134,30 @@ const Show = () => {
     const index = ONEUtil.timeToIndex({ effectiveTime })
     const neighbors = ONE.selectMerkleNeighbors({ layers, index })
     const neighbor = neighbors[0]
+
+    // Check for one address
+    let verifiedTransferTo = transferTo
```
polymorpher, line 139: Can be merged with above using utility function

givp: For sure, too much duplication. Will fix.

polymorpher:
https://user-images.githubusercontent.com/913871/123887877-20030780-d907-11eb-9019-1d57f53b3762.mov

givp:
@polymorpher please check again. I've moved the address checker to utils for both wallet creation and transfers. Added a message for the user when address check fails.

Also, I noticed the `transferAmount` check wasn't working since it's a BN so I fixed that [here](https://github.com/polymorpher/one-wallet/pull/21/commits/e7dc8ca4929270fbaa2efbb1d93e6527da5f48ef#diff-c82392b7eb3944456382eb45c15c56623a20b8043befea6d67055eeca24ba7baR116).

code/client/src/pages/Create.jsx
```diff
@@ -115,7 +123,7 @@ const Create = () => {
         t0: effectiveTime / WalletConstants.interval,
         lifespan: duration / WalletConstants.interval,
         slotSize,
-        lastResortAddress,
+        validLastResortAddress,
```
polymorpher, line 126: This won't work because it would change the key of the object.

code/client/src/pages/Create.jsx
```diff
@@ -125,7 +133,7 @@ const Create = () => {
         root: ONEUtil.hexView(root),
         duration,
         effectiveTime,
-        lastResortAddress,
+        validLastResortAddress,
```
polymorpher, line 136: Same issue as above

code/client/src/pages/Create.jsx
```diff
@@ -106,6 +107,13 @@ const Create = () => {
       message.error('Cannot deploy wallet. Error: root is not set.')
       return
     }
+
+    // Ensure valid address for both 0x and one1 formats
+    const validLastResortAddress = util.validateAddress(lastResortAddress)
+    if (!validLastResortAddress) {
+      return
```
polymorpher, line 114: no error message?

code/client/src/pages/Create.jsx
```diff
@@ -273,7 +281,7 @@ const Create = () => {
           <Heading>You are all set!</Heading>
           <Space direction='vertical' size='small'>
             <Hint>Wallet Address</Hint>
-            <Text>{address}</Text>
+            <Text>{address && getAddress(address).bech32}</Text>
```
polymorpher, line 284: The implementation of `getAddress` seems to be quite slow. The first thing it does seems to be a regex matching https://github.com/harmony-one/sdk/blob/3c7061887f25c6ba8def5af05672cdc4f44eb2dc/packages/harmony-utils/src/validators.ts#L6 , followed by multiple passes of the 20 bytes array and many string operations and comparisons.  It may be too slow to be in the rendering loop. We might need another state variable for it, watched by useEffect and only converted when it is at full length.

code/client/src/pages/List.jsx
```diff
@@ -7,12 +7,14 @@ import util from '../util'
...
 import Paths from '../constants/paths'
 import BN from 'bn.js'
+import { getAddress } from '@harmony-js/crypto'
 const { Text, Title } = Typography
 
 const WalletCard = ({ wallet }) => {
   const history = useHistory()
   const location = useLocation()
   const { address, name } = wallet
+  const oneAddress = getAddress(address).bech32
```
polymorpher, line 17: Same as above. Another approach is to useMemo

code/client/src/pages/Create.jsx
polymorpher: Also, this goes into the state model and persistent storage. It should not use bech32 address

code/client/src/pages/Create.jsx
polymorpher: Similar to the comment below, bech32 address should not be used here

code/client/src/pages/Create.jsx
```diff
@@ -106,6 +107,13 @@ const Create = () => {
       message.error('Cannot deploy wallet. Error: root is not set.')
       return
     }
+
+    // Ensure valid address for both 0x and one1 formats
+    const validLastResortAddress = util.validateAddress(lastResortAddress)
```
polymorpher, line 112: The bech32 address which is deemed valid here must be converted to ETH address prior to processing below

code/client/src/pages/Show.jsx
```diff
@@ -60,6 +62,7 @@ const Show = () => {
   const price = useSelector(state => state.wallet.price)
   const { formatted, fiatFormatted } = util.computeBalance(balance, price)
   const { dailyLimit, lastResortAddress } = wallet
+  const oneLastResort = lastResortAddress && getAddress(lastResortAddress).bech32
```
polymorpher, line 65: Same as above

code/client/src/pages/Show.jsx
```diff
@@ -38,6 +39,7 @@ const Show = () => {
   const wallets = useSelector(state => state.wallet.wallets)
   const match = useRouteMatch(Paths.show)
   const { address, action } = match ? match.params : {}
+  const oneAddress = getAddress(address).bech32
```
polymorpher, line 42: Same as above

code/client/src/pages/Show.jsx
```diff
@@ -131,11 +136,18 @@ const Show = () => {
     const index = ONEUtil.timeToIndex({ effectiveTime })
     const neighbors = ONE.selectMerkleNeighbors({ layers, index })
     const neighbor = neighbors[0]
+
+    // Ensure valid address for both 0x and one1 formats
+    const validAddress = util.validateAddress(transferTo)
+    if (!validAddress) {
+      return
```
polymorpher, line 143: No error message?

code/client/src/pages/Show.jsx
```diff
@@ -131,11 +136,18 @@ const Show = () => {
...
+    }
+
     const { hash: commitHash } = ONE.computeTransferHash({
       neighbor,
       index,
       eotp,
-      dest: transferTo,
+      dest: validAddress,
```
polymorpher, line 150: Address needs to be converted to ETH form

code/client/src/pages/Show.jsx
```diff
@@ -162,7 +174,7 @@ const Show = () => {
           neighbors: neighbors.map(n => ONEUtil.hexString(n)),
           index,
           eotp: ONEUtil.hexString(eotp),
-          dest: transferTo,
+          dest: validAddress,
```
polymorpher, line 177: Address needs to be converted to ETH form

code/client/src/util.js
```diff
@@ -71,6 +72,32 @@ export default {
     const fiat = (price || 0) * parseFloat(ones)
     const fiatFormatted = exports.default.formatNumber(fiat)
     return { balance, formatted, fiat, fiatFormatted, valid: true }
+  },
+
+  validateAddress: (address) => {
```
polymorpher, line 77: I see error message is prompted in this function instead of at each component themselves. And this function actually convert the address to ETH form, not just validate an address. Perhaps it should be renamed to `normalizeAddress` or something similar, and the UI prompt should be moved out to another function (e.g. `promptBadAddress`)

code/client/src/pages/Show.jsx
```diff
@@ -131,11 +136,18 @@ const Show = () => {
     const index = ONEUtil.timeToIndex({ effectiveTime })
     const neighbors = ONE.selectMerkleNeighbors({ layers, index })
     const neighbor = neighbors[0]
+
+    // Ensure valid address for both 0x and one1 formats
+    const validAddress = util.validateAddress(transferTo)
```
polymorpher, line 141: normalizedAddress?

code/client/src/pages/Create.jsx
polymorpher, line 112: normalizedAddress?

code/client/src/pages/Create.jsx
polymorpher, line 114: It would be quite hard to keep tracking error handling if the message prompt is absorbed in above function and its name does not suggest it does handle errors.

givp:
@polymorpher sorry for the confusion. `util.validateAddress` name is confusing as you suggested. It doesn't just validate, it confirms the format of one1 and 0x addresses but always returns ETH format. So `normalizeAddress` makes more sense. I'll change that. But the relayer calls are not affected. I'm still sending 0x addresses for all transactions, even if the user enters an one1 address.

In terms of console errors, I will add them in the function but was trying to avoid console noise .

givp:
This should be in a good place to merge. I'm going to revisit the `getAddress` speed issue once I've wrapped up some of the other tasks.

code/client/src/pages/Create.jsx
```diff
@@ -115,7 +123,7 @@ const Create = () => {
         t0: effectiveTime / WalletConstants.interval,
         lifespan: duration / WalletConstants.interval,
         slotSize,
-        lastResortAddress,
+        normalizedAddress,
```
polymorpher, line 126: This is still an issue. The expected object key is `lastResortAddress `

code/client/src/pages/Create.jsx
```diff
@@ -125,7 +133,7 @@ const Create = () => {
         root: ONEUtil.hexView(root),
         duration,
         effectiveTime,
-        lastResortAddress,
+        normalizedAddress,
```
polymorpher, line 136: same as above

givp: Ah apologies, fixing.

polymorpher:
I made a quick update on separation of error handling logic, and simplified the normalization: https://github.com/polymorpher/one-wallet/pull/21/commits/d9a92a586e74e1437645571d58e15e9413ad7ee1

PR #29, opened 2021-06-30
Adding wallet links to Explorer

givp:


givp:
@polymorpher I didn't link the address on the List cards since the whole card clicks through the Show page but let me know what you think.

polymorpher:
It looks good. I made a few quick changes to simplify things https://github.com/polymorpher/one-wallet/pull/29/commits/069d102b79629d551e389e9eedd211350cded1e5

polymorpher:
I also made a few improvements to resolve one (bech32) address and hex (checksum) address and to add general support for both everywhere in the ui including the URL

PR #33, opened 2021-07-02
Added Sentry.io

givp:
Added default Sentry.io SDK. `NODE_ENV` must be set to `production` to enable SDK. I will DM you the Sentry DSN credentials to go in the env `SENTRY_DSN`.


code/client/src/index.js
```diff
@@ -8,6 +8,16 @@ import { Provider } from 'react-redux'
 import { ConnectedRouter } from 'connected-react-router'
 import * as serviceWorker from './serviceWorker'
 import { initAPI, initBlockchain } from './api'
+import * as Sentry from '@sentry/react'
+import { Integrations } from '@sentry/tracing'
+
+if (process.env.NODE_ENV === 'production') {
```
polymorpher, line 14: environment variables need to be injected by webpack, which is done by EnvironmentPlugin and already configured. An variable already exists for `DEBUG` so this line can use that. `DEBUG` is already loaded in `config.js` but not yet exported so it would be more organized to load it there and import and refer to `config` here

code/client/src/index.js
```diff
@@ -8,6 +8,16 @@ import { Provider } from 'react-redux'
...
+  Sentry.init({
+    dsn: process.env.SENTRY_DSN,
```
polymorpher, line 16: Let's move all environment environment variable loading to `config.js`

PR #37, opened 2021-07-06
Added Sentry for capturing exceptions

givp:


polymorpher:
I will test and resolve merge conflicts later

givp:
> I will test and resolve merge conflicts later

No worries, conflict fixed.

polymorpher:
There were some major update on Show component logics (and all operations on the wallet), so I did more conflict resolution. Let me review exception capturing points more carefully later tomorrow. Exceptions alone can alert us but not very informative. We would need to capture the context of what the user is trying to achieve but failed. But we can start with exception capturing.

polymorpher:
Looks good. I removed the error capturing for safe functions in utility, since the mark "safe" basically is meant to be an intent to ignore errors or to handle it in a custom way. I also moved the error capturing inside Create to the general api error handler. As for the error capturer inside api and its init - I think the api calling functions themselves shouldn't be handling or capturing the errors. The callers should. Recently I moved the init functions to `code/lib`, where it is probably not a good idea to have Sentry. We can capture the init errors from outside inside `client/index`. Since there may be more than one errors in the init process, we can return the errors in an array as a separate return argument. In any case, let's do it later.

PR #41, opened 2021-07-06
Giv/32 display wallet stats

givp:
image


code/client/src/components/SiderMenu.jsx
```diff
@@ -27,16 +32,27 @@ const SiderMenu = ({ ...args }) => {
...
     history.push(Paths[key])
   }
+  const [stats, setStats] = useState(null)
+
+  useEffect(() => {
+    async function getStats () {
+      const statsData = await util.getWalletStats()
+      setStats(statsData)
+    }
+    getStats()
```
polymorpher, line 42: need to do setInterval here and clearInterval on exit, since a lot of people just leave the browser tab open and this would only run once

code/client/src/util.js
```diff
@@ -133,6 +134,21 @@ export default {
       return null
     }
     return parsedOtp
+  },
+
+  getWalletStats: async () => {
```
polymorpher, line 139: The response parsing logic can be moved inside api

code/client/src/util.js
```diff
@@ -4,6 +4,7 @@ import { isInteger, values } from 'lodash'
 import ONEUtil from '../../lib/util'
 import { AddressError } from './constants/errors'
 import config from './config'
+import api from './api'
```
polymorpher, line 7: Let's minimize the dependencies in util since many components may depend on this, including api itself

code/client/src/util.js
```diff
@@ -133,6 +134,21 @@ export default {
       return null
     }
     return parsedOtp
+  },
+
+  getWalletStats: async () => {
+    try {
+      const response = await api.walletStats.getStats()
+      const bi = BigInt(response.totalAmount) / BigInt(10 ** 14)
```
polymorpher, line 142: The responses should be parsed using bn.js and conversion from wei back and forth is already provided in util or ONEUtil

code/client/src/components/SiderMenu.jsx
```diff
@@ -27,16 +32,27 @@ const SiderMenu = ({ ...args }) => {
...
       <Row justify='center'>
-        <Link target='_blank' rel='noreferrer' href='https://harmony.one/'>
+        <Link target='_blank' rel='noopener noreferrer' href='https://harmony.one/'>
           <Image preview={false} src={collapsed ? HarmonyIcon : HarmonyLogo} style={{ cursor: 'pointer', padding: collapsed ? 16 : 32 }} onClick={() => history.push('/')} />
         </Link>
       </Row>
-      {!collapsed && <Text><Link href='https://harmony.one/1wallet' target='_blank' rel='noreferrer'>{config.appName} {config.version}</Link></Text>}
-      {!collapsed && <Row justify='center' style={{ marginBottom: 10 }}><Tag color='#0094c0'>Beta</Tag></Row>}
+      {!collapsed && <Text><Link href='https://harmony.one/1wallet' target='_blank' rel='noopener noreferrer'>{config.appName} {config.version}</Link></Text>}
+      <Row justify='center'><Stats>{stats && stats.count.toLocaleString()} wallets</Stats></Row>
```
polymorpher, line 54: This needs to be fail safe. If stats is null (for example when API server is down) this will show up as "undefined wallets"

code/client/src/components/SiderMenu.jsx
givp: @polymorpher I thought about adding this but the stats are only updated once every 24 hours on the backend.

code/client/src/components/SiderMenu.jsx
```diff
@@ -18,6 +18,11 @@ const Text = styled.p`
     color: #fafafa;
   }
 `
+
+const Stats = styled.p`
```
polymorpher, line 22: Can use Text

givp:
Updated.

polymorpher:
I made some style changes based on how npm shows stats, to make it less bland and consistent with the logo's branding colors.

image

givp:
Perfect

polymorpher:
image

PR #44, opened 2021-07-08
Hid relayer from topnav + tweaked stats

givp:
I had to tweak the stats because it did not fit for longer digits. I also commented out relayer options in topnav.

image

### DIFF
--- a/code/client/src/components/SiderMenu.jsx
+++ b/code/client/src/components/SiderMenu.jsx
@@ -1,6 +1,6 @@
 import React, { useState, useEffect } from 'react'
 import { useHistory, useRouteMatch } from 'react-router'
-import { Layout, Image, Menu, Row, Typography, Tag, Divider, Tooltip } from 'antd'
+import { Layout, Image, Menu, Row, Typography, Tag, Divider } from 'antd'
 import { PlusCircleOutlined, UnorderedListOutlined, HistoryOutlined } from '@ant-design/icons'
 import HarmonyLogo from '../assets/harmony.svg'
 import HarmonyIcon from '../assets/harmony-icon.svg'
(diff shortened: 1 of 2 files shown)


polymorpher:
Looks great. I think `1Wallet` is a bit redundant. `Total` too since it is implied by `Global`.  

We should still provide a way to let users set Relayer of their own choice (currently, via pasting a custom URL), since the user will be forwarding the user's proof for each transaction to the Relayer, and the Relayer would be the first (non-user) party to observe the proof. If the Relayer is offline, the user's transactions may be delayed or blocked. If the Relayer is hacked, until we make the security patch in the TODO, the attacker may use the revealed secret to steal funds. Let's make it possible to change the visibility by a toggle.

givp:
Good point, I'll make a lowkey toggle for the relayer - hidden by default.

givp:
Probably not the most elegant solution but OK for now. I will move it into a separate component later.

Screen Recording 2021-07-08 at 09 46 19 PM

polymorpher:
I made some updates. I hope you don't mind:
1. The global usage statistics: I spent about 30 minutes thinking about the design and trying different styles. First, I think the new layout (top-bottom stacking) is too distracting to the users. There is much too much text, and the tooltip does not give much extra information to the user. It is too vague on the methodology used to calculate the statistics. It may be helpful if we create a wiki page where we describe the steps we have taken to calculate the statistics, and link the tags there. Secondly, I think the color combinations (grey, light blue + white label text) are probably not the best choice. The white text is stealing away the focus from "Harmony" and "ONE Wallet", which are also bright white color. But the labels are apparently not as important as "Harmony" and "ONE Wallet". The grey-blue combination is not consistent with Harmony's logo (lightblue-teal gradient). Thirdly, I think it is taking too much space. The wallet statistics are not particularly helpful to the user, so it should not take the top portion of the space, or to occupy more space than the main buttons which lead to three main features of the wallet.
2. I simplified the structure around the relayer toggle you added.

PR #45, opened 2021-07-12
CLI for SlowMist audit 

polymorpher:
No contract code was changed. The core library in `/code/lib/` was expanded with some shared configurations, API invocations and data processing, and general workflows around using the APIs.


polymorpher:
#37 #41 #44 were also merged

PR #54, opened 2021-07-19
use totp protocol to open google authenticator

haolinj:


polymorpher:
It works fine on iOS. Turns out Android Google Authenticator can't interpret the QR code with `totp://` scheme

haolinj:
> It works fine on iOS. Turns out Android Google Authenticator can't interpret the QR code with `totp://` scheme

Good find, it was a long shot on this. We may need to think about something else to make it open only for Google Authenticator.

polymorpher:
Let's revisit this when we have a more complete solution, especially for Android.

PR #55, opened 2021-07-20
Clear and focus otp input on error and various improvements on post-transfer experience

givp:
The only place I could find where clearing and focusing made sense was on wallet creation. Transfer check happens later and it seems inputs are cleared anyway. Restore has no user input from what I can tell?

`react-otp-input` doesn't seem to have a method for focusing to the first input so I had to do it with vanilla JS selectors.


polymorpher:
The standard way to control the inner components is to use references. I added some updates as an example. The methods available to the component are in https://github.com/devfolioco/react-otp-input/blob/main/src/lib/index.jsx

PR #57, opened 2021-07-25
ONEWallet broken down into two classes - separating the logic of toke…

ivan-homoliak-sutd:
OneWallet broken down into 2 contracts to isolate the logic of token tracking and simplify the audits. All the former functionality and API is preserved.


hashmesan:
I’m not sure token tracking is necessary because the the api server can index when it process each block.

polymorpher:
> OneWallet broken down into 2 contracts to isolate the logic of token tracking and simplify the audits. All the former functionality and API is preserved.

Looks good. I will do more end to end testing later today

polymorpher:
> I’m not sure token tracking is necessary because the the api server can index when it process each block.

The contract should maintain the list internally, since these tokens are supposed to be recoverable to recovery address. Another issue is that, the tracking by an external server has been prone to problems in practice such as delay, non-uniform standards, and undocumented methodologies. In particular, in other NFT projects I work on, I found many wallets are depending on OpenSea APIs, such as Coinbase Wallet and Trust Wallet, DeFi wallet, and others. OpenSea's tracking is delayed by at least a few minutes. OpenSea's API also pull contract information only once. It also caches imagery metadata, store and downsample them on centralized servers  (Google Storage), and serve only the cached versions and URLs via API. The APIs also often get the names wrong, and are using their own undocumented standards to retrieve names. If we are to rely on similar tracking methods, even with our own APIs, we (or developers relying on us) may run into the same problems. Not to mention the upfront and ongoing cost to develop and maintain such server and APIs.

hashmesan:
The issues you are referring may be bugs, short-falls of specific implementations.  Of all the blockchains I've worked on, I've observed a common design pattern that is similar to[ memory-storage architecture](https://hazelcast.com/wp-content/uploads/2019/08/diagram-MemoryCache.png):

\  application-cache data /
  \     indexed data        /
     \  on blockchain /

Bitcoin's account transactions are indexed on the client database.
Ethereum transactionCount, transactions hash, etc
Harmony - Client & API -- HRC20/721/1155 tokens shows on explorer: https://explorer.harmony.one/address/0x2b73feb5af7ef568a194499da1bce71ff35cb209

I'm not sure other specific constraints in your implementation I may be missing. But I much prefer to leverage the existing infrastructure for tracking. To me it's much cheaper to develop / maintain on API than on blockchain, and it cost less tx fee to deploy & maintain.

polymorpher:
The explorer fails to show tokens in my wallets that comply with EIP-721 and 1155 standards: https://explorer.harmony.one/address/0x6c01e9fd32da6afc9673d8aeb795be320e0d2961

Until the infrastructure meets a minimum standard such that it does what it purports to do within reasonable expectations, it cannot be relied upon. Using a few bytes to track what each wallet does not introduce burden to the blockchain, nor any noticeable gas expenditure, especially after the first time the wallet receives or sends any token. If the extra bytes becomes burdensome for the blockchain, the blockchain itself should be substantially improved and revised. The burden should not be shifted to application developers.

polymorpher:
No issues discovered during manual end-to-end tests

quocble:
@polymorpher I agree - it cannot be relied on. I am running into the same issue as you.   It's written in nodejs /async.. I wrote one for Quanta blockchain in Golang with sync & indexing and it's hard to get it right...and it has to be done in sync Golang/C++/Python so that it is predictable, and atomic transactions.  I'm raising this issue -- a lot of developers will benefit from this.

hashmesan:
@polymorpher I agree - it cannot be relied on. I am running into the same issue as you. It's written in nodejs /async.. I wrote one for Quanta blockchain in Golang with sync & indexing and it's hard to get it right...and it has to be done in sync Golang/C++/Python so that it is predictable, and atomic transactions. I'm raising this issue -- a lot of developers will benefit from this.

PR #58, opened 2021-07-26
Wiki Review: Client Security

polymorpher:


wiki/Client Security.md
```diff
@@ -0,0 +1,95 @@
...
+> 
+> Since the OTP is confined to a 6-digit numerical number, and the generation process relies on SHA256, the right OTP (and by extension, its EOTP) at any given time can be easily enumerated by brute-force, for any given leaf hash value. See for the code of doing so at 
+> 
+> https://github.com/polymorpher/one-wallet/blob/afe39e657df1522ee8941e386025c4ddafc3ab94/code/lib/onewallet.js#L141 
+
+An adequate solution must make it exceedingly difficult, if not practically impossible, for any party to perform any transfer (or similar operations) without the input from the Google Authenticator. At the same time, it must not significantly slow down the wallet creation process, or the process of making an authorized operation with the correct authenticator code.
+
+This leaves us several options to enhance the security:
+
+1. Substantially increase the search space required for brute force attack, from the current 1 million possible OTPs.
```
polymorpher, line 19: By @ivan-homoliak-sutd :
```
I do not think this will help since the attacker who already has the seed (i.e., its hash) can bruteforce just one particular EOTP in the near future. She does not need to make bruteforce across multiple EOTPs to significantly increase her chances of success. Maybe you mean the scrambled design, but even here, attacker might do some tricks to get EOTPs (see below)
```

wiki/Client Security.md
```diff
@@ -0,0 +1,95 @@
...
+
+If we choose µ=10<sup>6</sup>, the attacker would need compute 10<sup>12</sup> hashes to brute-force both the randomly generated number and the OTP. We know µ=10<sup>6</sup> is a practical number for runtime, since in ONE Wallet v0.1 we perform this number of hashes every time we generate a new wallet, and it is typically completed within less than a few seconds in the client's browser. 
+
+Using the [hashcat benchmark](https://github.com/siseci/hashcat-benchmark-comparison/blob/45a27b32a2f24d317cc29741d64fc739f3a30cb5/1x%20Gtx%201080%20TI%20with%20Overclock%20Hashcat%20Benchmark) for reference, a highend NVIDIA GPU would enable the attacker to complete the brute-force in roughly 10 minutes. However, if we require the user to provide two OTPs inputs, the attacker's job would become practically impossible: now he would need 19 years to complete the bruteforce using a highend NVIDIA GPU. The attacker would need to design an ASIC on par with the performance of AntMiner in Bitcoin mining (~100TH/s) to complete the bruteforce in a reasonable amount of time. At ~100TH/s, the time required is ~3 hours.
```
polymorpher, line 53: By @ivan-homoliak-sutd : 

> You further mention that user has to provide 2 OTPs - I agree, more entropy is always better to protect against bruteforcing. However, the problem of using TOTPs is that the user has to wait (from 5 - 25 sec) to get the next TOTP from GA, which deteriorates the usability. So, it'd be good to consider the transferred amount and use 2 TOTPs when you are over a threshold, e.g., 100USD (Depending on the cost of the attack bruteforcing a single TOTP + controlled randomness)

wiki/Client Security.md
```diff
@@ -0,0 +1,95 @@
...
...ng this intuition is to store each leaf at the position of the OTP they correspond to and a pointer to their neighbor. Unfortunately this naive approach leaks the OTP code of each leaf and renders OTP useless. A mild improvement from the naive approach is to swap each leaf with its neighbor's position (after first repositioning all leaves to their OTP values) and remove the pointer to the neighbor (since neighbor is part of the Merkle Proof, not the leaf node itself). This construct no longer leaks the OTP for each leaf, but an attacker would still be unable to identify the location for each pair of leaves by pre-comptuing the parent hash of all pairs of leaves, which can be done in M<sup>2</sup> operations and would only require 32 bytes * M amount of memory. Since each operation is a SHA256 hash, based on the analysis in Section I, an attacker with a high-end GPU only needs ~10 minutes to identify exact positions of all leaves before they are scrambled. Moreover, we cannot use the techniques in Section II by replacing the hash function to a complex one, since the hash function must be supported on blockchain as well and must be economical and fast enough to compute on blockchain.
```
polymorpher, line 74: From @ivan-homoliak-sutd :

> But Merkle proof is computed at the client. So, with scrambled design, I do not see how you can compute it since you do not know where are the other leaves required to compute the proof. See also below. (This is similar even in the improved version.)

> I understood that you store only leafs and not intermediary nodes of Merkle tree at the client. However,  it is also convenient to store intermediary nodes, so you can construct proofs at the client faster - just by accessing the continuous array of leafes and intermediary nodes (no need to do Merkle tree aggregation/reduction). This holds only if you do not have scrambled memory. With scrambled memory, this might be a problem. Anyway, I do not understand how the client constructs proofs even with scrambled design. To do so, the client needs also the knowledge of positions of other leafs (additionally intermediary nodes in the tree), while these positions can be obtained only when other OTPs are provided (which is unrealistic). Could you please clarify?

wiki/Client Security.md
```diff
@@ -0,0 +1,95 @@
...
+
+With this approach the number of operations required for simple brute-force is increased to (4M)<sup>8</sup> since the attacker would have to enumerate all possible combinations of chunks for each pair of neighbors and test whether they produce the required hash value in a any parent node. There may be more efficient brute-force attacks which we have not explored or analyzed.
+
+Overall, this techniques is the least straightforward, but may be the most secure and the fastest for runtime execution. In any case, this technique can be composed with the techniques in Part I or II or both to obtain multiplicative security protection, since they operate at different part of the process.
```
polymorpher, line 93: From @ivan-homoliak-sutd : 

> You do not assume intermediary nodes of Merkle tree already submitted to blockchain within some old proofs. The attacker is able to permutate, I'd guess, up to 4 or 8 tuples of EOTPs from the client's memory and check it against some future intermediary node (at the level 16 or 17 from the root) submitted in some Mrekle proof. So, in theory, the attacker has 8*30s = 240sec to find the correct permutation (assuming that he can do max. 8 touples, maybe she can do more and will have more time).

wiki/Client Security.md
```diff
@@ -0,0 +1,95 @@
...
+
+The choice of **M** is limited by the number of digits supported by Google Authenticator. In Android, this is unfortunately fixed to 6 at this time. 
+
+The choice of **N** is set by the lifespan of the wallet and the number of operations per time interval. The time interval duration is fixed at 30 seconds at this time due to limitation of Android Google Authenticator. The default lifespan of a wallet is 1 year, which leads to 3600 * 24 / 30 * 365 = 1051200 ~= 10<sup>6</sup> leafs.
+
+We also denote the hash of the OTP Seed as *k<sub>h</sub>* and the OTP Seed itself as *k*, similar to the notation in [Protocol](https://github.com/polymorpher/one-wallet/blob/43197c70e36cb58c2884c423c1e665feff232042/wiki/protocol.pdf).  
+
+As described in Protocol § 1.1.5, each leaf node at index *i* is computed by the following process, which we will denote as ** GenLeaf**:
+
+1. Taking first 26 bytes of *k<sub>h</sub>*, and concatenate with 2 bytes of nonce, and 4 bytes of OTP corresponding to index *i*. All bytes are in big endian encoding.
```
polymorpher, line 37: From @ivan-homoliak-sutd  : 

> What is nonce, and who generates and stores it? If it is stored by the client, then to me it means that it is the same as the h_k

wiki/Client Security.md
```diff
@@ -0,0 +1,95 @@
...
+
+After comparing several alternatives, we believe [argon2](https://github.com/P-H-C/phc-winner-argon2) is the optimal choice to replace step (2) of **GenLeaf** in the next version of ONE Wallet. Similar to scrypt, it can be configured to use a large size of memory and impose an arbitary workload to slow down the computation. Aside for its multiple advantages over other candidates (which are discussed at length in their Github repository, the password-hashing competition, and their paper), here are the other factors we have considered:
+
+- Argon2 has a [browser-based implementation](https://github.com/antelle/argon2-browser) based on WebAssembly. It offers satisfactory performance under reasonably low memory settings, and can complete within a reasonable amount of time for 10<sup>6</sup> repeatitions. Therefore, it would not slow down the setup process
+- Based on our knowledge, no one has built an ASIC miner for computing Argon2 hashes at this time.
```
polymorpher, line 65: From @ivan-homoliak-sutd  : 

> This might change in the future, especially if OneWallet will be successfull :) What about to employ verifiable delay functions (VDFs), such as recursice hash computations (e.g., https://en.wikipedia.org/wiki/PBKDF2) instead of scrypt or on top of it?

wiki/Client Security.md
polymorpher: I should clarify in the article that the search space here is within the context of one EOTP. To brute-force, the attacker needs to enumerate all possible inputs to and check which one generates the target EOTP. Since the attacker is assumed to have already obtained all information at the client, the input space is left with only 1 million possibilities: the OTP from 0 to 100000, which the attacker needs to enumerate

wiki/Client Security.md
polymorpher: The nonce is a counter per time interval that is visible on the contract, to enable multiple operations per time interval. Since it is public, it needs not to be considered for the purpose of security strength analysis here. The nonce is implemented, but not documented at this time (not a priority). I will document it in the next major revision of the Wiki.

wiki/Client Security.md
polymorpher: The proposal I had in mind is to generate another seed and provide another QR code for the user to scan. Afterwards, we regenerate the Merkle Tree using combined OTPs, and use a combined hashed seed (by taking half of each hashed seed). We will need to replace the acting root in the contract during the process. At run-time, the two OTPs would be simultaneously visible to the user, so the user needs not to wait.

wiki/Client Security.md
polymorpher: This sounds promising

wiki/Client Security.md
polymorpher: The entire tree is stored at the client, not just the leaves. I don't think there is much benefit of storing only the leaves since it would only reduce storage consumption by 50%.

wiki/Client Security.md
polymorpher: Denote the leaves as nodes at level 0. It is assumed that correct, unscrambled parent nodes (level 1) are visible to the attacker at the client side. Previous leaves submitted to the blockchain are also visible to the attacker. However, to find the correct permutation of chunks for the current leaf, the attacker still needs to enumerate over a large space

wiki/Client Security.md
ivan-homoliak-sutd: So you do not assume that the attacker might also try to brute-force other EOTPs, whose hashes (or hashes of their near parents) she might obtain from the blockchain? I do not see the reason why it should not be considered. Can you outline?

wiki/Client Security.md
ivan-homoliak-sutd: Ok, this sounds better, though it might be a bit unclear for the user to keep two OneWallet entries in Google Auth. She might accidentally drop one due to suspicion of duplicity.

polymorpher: That shouldn't be an issue. We can set arbitrary names (issuer, account) in Google Authenticator's UI.

wiki/Client Security.md
polymorpher: They can brute-force other EOTPs but they can only obtain past EOTPs - which don't seem to be useful to brute-force?

wiki/Client Security.md
ivan-homoliak-sutd: Yes, I agree to store the entire tree. 

"the number of operations required for simple brute-force is increased to (4M)^8" 
=> 4M does not seem correct. When you generate the 4 random numbers by PRNG, you can put the range of even 2^256 (by using hash) instead and store the elements in the mapping, and according to your formula, it'd mean unbelievable (2^256M)^8. However, the entropy of one OTP is only 2^20 ~ 2.5B, which, to me, is also manifested to PRNG you use. So, in fact, you have only 1M options => (1M)^8 ~ 160bits; which is pretty good.

Assuming that the left sibling was already submitted, this degrades to (1M)^4 ~ 80bits, which is still strong enough to be brute-forced in 30s or 60s (until the expiration of the right sibling).

**Some minor comments:**

Also, can you please write that by the "result", you mean leaves (further containing controlled randomness)?

"...attacker would still be **unable** to identify the location" => able

"Moreover, we cannot use the techniques in Section II by replacing the hash function to a complex one, since the hash 
function must be supported on blockchain" 
=> The complex hash function would be used only for computation of EOTP at the client, which is agnostic to the blockchain; so it can be used.

"to generate 4 **consecutive** numbers between [1, 4M]" 
=> why consecutive? I'd say random.

wiki/Client Security.md
ivan-homoliak-sutd: OK, it is seems fine. Thanks.

wiki/Client Security.md
ivan-homoliak-sutd: OK, thanks.

ivan-homoliak-sutd: BTW, I do not understand how there can be **multiple valid operations** per a single time interval. Intuitively, only a single valid operation can be submitted at the time since the user can see only a single TOTP in GA. Even if there were submitted two, the reveal stage would finish only the first one.

wiki/Client Security.md
ivan-homoliak-sutd: Ok, got it.

wiki/Client Security.md
ivan-homoliak-sutd: Ok then, but keep it in mind that this needs to be investigated during some alpha testing / usability studies.

wiki/Client Security.md
polymorpher: For each time interval, we encode `maxNumOperationsPerInterval` number of different EOTPs during setup by inserting different nonces at the input. See https://github.com/polymorpher/one-wallet/blob/1ed4417b9ba7c6d6cf57b3a0a311023fca800b9a/code/lib/onewallet.js#L25. The smart contract keeps track of the expected nonce for each time interval, and increment the nonce by one upon a success reveal. The client gets the expected nonce from the smart contract (or simply use 0, if `maxNumOperationsPerInterval=1`) and use that as part of the input in addition to the OTP for the current time interval. Since only EOTP is revealed, an attacker would not be able to know the true OTP, hence cannot use the nonce to forge an operation.

wiki/Client Security.md
polymorpher: 1. Yes, **unable** is a typo and should be replaced with **able**.
2. We cannot use the complex hash function because the brute-force operation being considered here is the one that hashes two leaf nodes into a parent node. The attacker only needs to recover the correct position of leaf nodes here from the scrambled memory layout. After correct positions are identified, the attacker can proceed to brute-force one or more individual leaf nodes, which the complexity is only additive, not multiplicative , to the overall cost.
3. Yes, `Result` as used in the opening sentence of this line `We can further improve the security of **GenLeaf** by placing the result in the memory` means a leaf (not leaves, since **GenLeaf** only generates one at a time).
4. The comment about consecutive numbers seem to be referring to a different line. Let's open another thread there because there are too many comments in this line.
5.  Similarly, the comment about number of operations required for simple brute-force is on a separate line. I will copy your comment and start a thread there.

wiki/Client Security.md
```diff
@@ -0,0 +1,95 @@
...
+
+These two naive approaches are unsatisfactory because they are missing a few critical elements:
+
+1. They did not use randomness.
+2. Their position mapping function for each leaf is invertible.
+3. They did not expand the search space for brute-force attacker (confined within M<sup>2</sup>)
+
+To address these issues, we introduce the following technique:
+
+1. For each leaf, we use the corresponding OTP as the seed for a random number generator, to generate 4 consecutive numbers between [1, 4M]. Note this mapping is deterministic (hence repeatable) but is not invertible.
```
polymorpher, line 84: From @ivan-homoliak-sutd :
> why consecutive? I'd say random.

wiki/Client Security.md
```diff
@@ -0,0 +1,95 @@
...
+
+With this approach the number of operations required for simple brute-force is increased to (4M)<sup>8</sup> since the attacker would have to enumerate all possible combinations of chunks for each pair of neighbors and test whether they produce the required hash value in a any parent node. There may be more efficient brute-force attacks which we have not explored or analyzed.
```
polymorpher, line 91: From @ivan-homoliak-sutd : 

> 4M does not seem correct. When you generate the 4 random numbers by PRNG, you can put the range of even 2^256 (by using hash) instead and store the elements in the mapping, and according to your formula, it'd mean unbelievable (2^256M)^8. However, the entropy of one OTP is only 2^20 ~ 2.5B, which, to me, is also manifested to PRNG you use. So, in fact, you have only 1M options => (1M)^8 ~ 160bits; which is pretty good.
> 
> Assuming that the left sibling was already submitted, this degrades to (1M)^4 ~ 80bits, which is still strong enough to be brute-forced in 30s or 60s (until the expiration of the right sibling).

wiki/Client Security.md
polymorpher: Should be rephrased to `to consecutively generate 4 random numbers between [1, 4M]`. The consecutive requirement is for determinism, so that a user can restore the entire wallet construction and scrambled memory layout using only the same OTP seed.

wiki/Client Security.md
polymorpher: ~4M comes from rough size of a dense array storing the chunks. For the array to be dense, its size should roughly be ~1M codes multiply by 4 chunks, i.e. 4M.  Even if we use a large range for the random number generator (e.g. 400M instead of 4M), it would be only as good as 4M because most positions would be zero. The result array would be sparse, and its equivalent dense representation would still only be roughly 4M long.   

Consider further, that if we generate more than 1M codes. Because there are only 1M OTPs in each 6-digit code, we would start to observe repetitions in chunks. The resulting dense array would still be effectively 4M in size for the reason that two or more repeating chunks are interchangeable, hence are only as good as one chunk and should be only counted as one entry in a dense array.

However, in a sparse representation where random range is set to a high value (again consider using 400M as an example), we may fill in the zero values with random noise. This construct may be effective and increase the number of operations required for simple brute-force to a very high value. More effective brute-force might exist, since we have not analyzed brute-force resistance of each component involved throughly (random number generator, whether patterns can be recovered from the scrambled layout, etc.). Note further that no matter how effective the scrambled technique is, we are limited to the 20-byte randomness in OTP Seed.

polymorpher:
Merging for now. Scrambled Memory Layout section needs some further work - it is effective to prevent ASIC and GPU attack because the whole scrambled layout has to be loaded in memory for each brute-force attempt. But it does not seem to increase the order of magnitudes of operations required, on top of the first two methods. The analysis given is reasonable for brute-forcing the position of the leaves. It is not the correct analysis for an attacker aiming to recover the OTP for a given timestamp. I will create a separate issue to document this.

wiki/Client Security.md
```diff
@@ -0,0 +1,95 @@
+## Client Security
+
+As noted in https://github.com/polymorpher/one-wallet/issues/5, in ONE Wallet v0.1 the authenticator cannot provide security protection if the client is compromised. A compromised client means the information stored on the client is leaked to a malicious third-party. In the current setup, the third-party may use these leaked information to easily reverse engineer (by brute force) the proof needed to perform an unauthorized transfer from ONE Wallet. The 6-digit code input from Google Authenticator would not be required.
```
ivan-homoliak-sutd, line 3: Probably the most important question that I forgot to ask before is whether is this attacker model even realistic? 

**Intuition:** If the attacker (e.g., malware) has access to the client, she will very likely be able to tamper with the client, and thus capture OTP from the user, while no brute-forcing is required. So please think about use cases in which this does not occur, while only leakage of client's data occur.

wiki/Client Security.md
ivan-homoliak-sutd: We clarified with Aaron today, and now I agree that it makes sense principally. We also clarified that there is no 1M OTPs, but 1M times the max value of nonce (currently 256). So the client generates 256 EOTP from a single GA OTP, which is further aggregated to the Merkle tree. My suggestion is to decrease the max number of the nonce to 8 or 16, since executing more operations within the same 30s interval is unrealistic even in the case of a shared wallet.

polymorpher: Correction: current max value of nonce is 1 (i.e. maxOperationsPerInterval) for all clients. The line of code referenced above says "2 bytes for nonce" only indicates that 2 bytes are reserved for nonce in the encoding procedure.

wiki/Client Security.md
polymorpher: Let's consider two attack patterns: (1) general large scale malware attack, and (2) an attack targeting a specific user for a specific purpose. It is possible that (1) may lead to (2) if the malware is undetected for a long period of time and the user, presumably a high value target, fails to act against the malware within this period of time. This scenario will be discussed separately. 

Furthermore, within each context, it is important to distinguish various levels of "compromised client" in real-world settings: (a) temporary control by adversaries for a brief window of time, (b) prolonged, limited read access by adversaries (mouse, keyboard input, and storage), (c) prolonged, unlimited, protected read access by adversaries (display is accessible by attacker, but sandboxed browser/app sessions are still protected),  (d) prolonged, unlimited, unprotected read access by adversaries. Within each of (a)-(d), it is also important to distinguish several levels of write access: (i) ability to obstruct client's network, (ii) ability to obstruct client's operation, (iii) ability to write to client's storage, (iv) ability to alter client's memory, (v) ability to alter client's code.

General operations under the wallet with Client Security should be secure against attackers under (a)(b)(c), even with write access (i)(ii)(iii)(iv). General operations do not include those exposing OTP Seed at the client (Create, Restore). For seed-related operations, the list is reduced to (a)(b) and (i)(ii)(iii), though some execution environments (including browsers) may protect a region in the display so (c) can also be included. 

Using modern browsers (user coverage >95%) and mobile phones, the situation of (d) and (iv), (v) are very unlikely, even if the malware gains a high level access in the user's system. The browsers (in conjunction with OS) and phone's operating systems do an excellent job in protecting memory access and the executable code. An attacker with root access would have better luck providing a "fake browser" shim and take control from there. That attack is easy to detect and browsers have various safeguards against that. Moreover, most devices made after 2016 have  secure enclave capability to provide a trusted execution environment. The trusted execution environment would make the execution safe (including memory access) against even attackers with root access. The support from OS is already implemented, and is available in at least macOS. The support from browser might take another 1-3 years to become generally available.

Thus, unintended leakage of stored credentials and keys to access them are the primary targets for attack, because there is simply no way the browser can protect from other programs intercepting keyboard inputs and reading from disk storage. For example, MetaMask stores a password-protected private key at a disk location: https://ethereum.stackexchange.com/questions/52658/where-does-metamask-store-the-wallet-seed-file-path#64038. A malware only needs to log the user's keyboard input and send a copy of the file at the disk location to the attacker to gain full access to the user's wallet.  A simple Google search would yield many detailed stories of how people lost their wallets due to this form of attack.

The analysis brings us back to the attack patterns in (1) and (2). The attack patterns in (1) generally rely on spotting vulnerable clients, intercepting necessary credentials, reporting credentials back to the attacker, similar to the MetaMask example above. It would be ineffective against our wallet for the reasons already explained in Client Security document. One could argue that the attack pattern can be modified and automated, such that the malware automatically intercepts the OTP(s), blocks the user's executions, and send funds to an alternative address before the user's normal execution is completed. This attack would inevitably require (d) (or (b)/(c) in some cases, discussed below) and at least (i) or (ii). On mobile, (d) is well protected by the OS and is extremely hard to achieve, as explained above. On desktop, (d) is less protected than mobile but the browser still offers good protection. In the case of (b) or (c), the attacker would have to identify and extract the OTPs in a sequence of keyboard inputs with timestamps. With the extra information provided in (c) and assumes the display region is not protected, the job becomes substantially easier. However, it can be disrupted by generating synthetic keyboard input noises and/or requiring the input to be made by mouse clicks on a protected, synthetic keypad with numbers placed at scrambled positions. Since the attacker only has limited time (60 seconds by default) to complete the attack (and very likely only one chance, since the longer it takes, the easier it is to detect), an automated program is very unlikely to succeed. The reason for time limit is as the following: in both (i) and (ii) the attacker could choose to disrupt only ONE Wallet related network connection to make the attack harder to detect. However, the attacker would have to preclude the user's client from committing the transaction altogether, as otherwise the attacker's commit-reveal would not succeed by the design of the mechanism (side note: we should require nonce==expectedNonce in the smart contract, instead of >=). This prevention of commit would at least raise suspicion. Moreover, the attacker has to complete his commit and reveal within a delay factor of REVEAL_MAX_DELAY, which is set to 60 seconds as of now. Assuming the attack succeeds, it would also be immediately noticed by the user.

An attacker following attack pattern (2) would follow a similar pattern described in the last paragraph, but exercise a greater control over whether, when, and how the attack is made and how user's legitimate transaction is blocked. It would also have the luxury to observe the keyboard and mouse input patterns (and potentially visual output from the display) and make far more intelligent and effective attempts than an automated program for the mass. Such attack has to be highly customized and carefully made. Nonetheless, it would not be able to prevent the user from noticing the attack and the attacker would not have a second chance after it is detected.  These rare, high-value, vulnerable individuals should consider utilizing Yubikey as proposed in #5 .

ivan-homoliak-sutd:
### General operations under the wallet with Client Security should be secure against attackers under (a)(b)(c), even with write access (i)(ii)(iii)(iv).

Regarding (b)(c), in which keyboard + display is accessible to the attacker - what if the attacker steals OTP from the screen or the keyboard? The attacker has write access to the network, so she can exfiltrate the OTP + proof to her server, where she can proceed further.

ivan-homoliak-sutd:
### The trusted execution environment would make the execution safe (including memory access) against even attackers with root access.

I am just thinking aloud. The goal of TEE is to isolate a local trusted application from the possibly untrusted OS, which enables to safely store sealed secrets and provable executing of local code upon a remote request  (so the remote party has proof that a correct code was executed). However, I do not think that trusted remote code execution is the use case for 1Wallet - we do not have any remote party that would rely on the code running within a local enclave. In the case of sealing use case, this might be related since we need to store the confidential h(seed) at the client, which can be sealed and accessed only through the code of enclave. This means that enclave would serve as the query oracle for the attacker (who can execute the client's code). Nevertheless, if the attacker (b) / (c) having (i)(ii)(iii)(iv) can intercept OTP from display or keyboard, she can query the enclave with the correct parameters and avoid any rate limit filters on accessing encrypted/scrambled proofs, while she can proceed in the attack.

polymorpher:
> ### General operations under the wallet with Client Security should be secure against attackers under (a)(b)(c), even with write access (i)(ii)(iii)(iv).
> Regarding (b)(c), in which keyboard + display is accessible to the attacker - what if the attacker steals OTP from the screen or the keyboard? The attacker has write access to the network, so she can exfiltrate the OTP + proof to her server, where she can proceed further.

Yes, it does look confusing. I need to revise that paragraph and add to Wiki. There are two cases here: system-level screen/keyboard access (e.g. key logger malware, or something at hardware level), and process-level (e.g. a javascript / chrome extension key listener). The process level stuff is easy to guard against - just trigger some bogus key presses. We can password-mask the OTPs to make screen read access ineffective. System-level keyboard logging without screen access is also not quite effective - it can be made difficult for the attacker to decipher which key presses correspond to OTP and which are not. System-level keyboard logging + screen read access at the same time is problematic. But attacker has limited time, has one chance only, and is easy to be detected - as analyzed before.

wiki/Client Security.md
ivan-homoliak-sutd: 2B => 2^16 = ~16k. Would you intend to use 1B instead? (which would enable 256 nonces/operations)

polymorpher: Yes but that may make operations less efficient, since a WORD is 16-bits and usually operations are done most efficiently on 4-bytes at a time (e.g. integer size). Having 1-byte in the middle of all these misaligns the rest of the stuff and make them positioned at odd bytes

polymorpher:
> ### The trusted execution environment would make the execution safe (including memory access) against even attackers with root access.
> I am just thinking aloud. The goal of TEE is to isolate a local trusted application from the possibly untrusted OS, which enables to safely store sealed secrets and provable executing of local code upon a remote request (so the remote party has proof that a correct code was executed). However, I do not think that trusted remote code execution is the use case for 1Wallet - we do not have any remote party that would rely on the code running within a local enclave. In the case of sealing use case, this might be related since we need to store the confidential h(seed) at the client (+ data for proofs), which can be sealed and accessed only through the code of enclave. This means that enclave would serve as the query oracle for the attacker (who can execute the client's code). Nevertheless, if the attacker (b) / (c) having (i)(ii)(iii)(iv) can intercept OTP from display or keyboard, she can query the enclave with the correct parameters and avoid any rate limit filters on accessing encrypted/scrambled proofs, while she can proceed in the attack.

Yes, TEE would only prevent attacker from tampering with the execution code. The attacker would have the same level of access to the enclave compared to the user. However, we can add a counter to the protected code in TEE such that any query would increment the counter. The attacker's actions would alert the user immediately.

Issue #59, opened 2021-07-27
DoS vulnerability in secure commit-reveal mechanism for core wallet major version 6


polymorpher:
Shortly after core wallet major version 6 is released (and overall version 0.4) in https://github.com/polymorpher/one-wallet/pull/56, Andrianna Polydouri and  her team from Common Prefix discovered a DoS issue:

> In the previous version a commit tx would require only one hash containing all the information for both merkle proof and operation's parameters. In the new version proof's hash and parameters' hash are separated and saved in the commitLocker  mapping under key  value hash. The reasoning behind this is that if an attacker tried to front-run upon victim's reveal tx, her attack would fail since she would try to commit under the same hash value, which is already committed to.
>
> The attacker cannot "steal" the amount that the victim would transfer anymore, however she can easily cause DoS for any user by front-running any commit transaction and "reserving" the commitLocker[hash] with arbitrary paramsHash.
> 
> Let's say that a user wants to execute a tx and he submits a commit tx (hash, paramsHash1) for this purpose. Then the attacker will front-run this tx and first submit her commit tx (hash, paramsHash2). paramsHash2 could be just garbage. The attacker's commit tx is executed first getting the commitment under hash reserved, so resulting in user's tx to fail (cannot commit to the same hash).
Now the user should submit another commit tx with a different hash' and the same paramsHash1, but the attacker can repeat the attack. And this can be repeated again and again. The attacker cannot steal funds but she can delay user's transactions, considering she is able to front-run his transactions, for as long as she wants.

I proposed the following solution, which appears to be effective upon a preliminary review

> You are right. Each commitHash should allow multiple entries, instead of just one.  Each paramsHash should be accompanied with another hash, which is hash(paramsHash . eotp) - call it verificationHash 
> 
> Since the attacker does not know eotp, it cannot generate a correct verificationHash. If the attacker wants to frontrun, while the real eotp is being revealed, the attacker may commit another value to commitHash with (paramsHash2, verificationHash2), with his own parameters.
> 
> We can prevent the frontrun by only allow executing the first entry that has a valid verficationHash. Since the user’s (paramsHash, verificationHash) is always commited before attacker’s (paramsHash2, verificationHash2), the attacker would be unable to get (paramsHash2, verificationHash2) executed.

PR #60, opened 2021-07-27
Resolve DoS vulnerability at the commit stage of the commit-reveal mechanism

polymorpher:
See https://github.com/polymorpher/one-wallet/issues/59


code/contracts/ONEWallet.sol
```diff
@@ -351,25 +393,40 @@ contract ONEWallet is TokenTracker {
...
+        }
+        revert("No valid commit");
     }
 
-    function _completeReveal(bytes32 commitHash) internal {
-        Commit storage c = commitLocker[commitHash];
-        require(c.timestamp > 0, "Invalid commit hash");
+    function _completeReveal(bytes32 commitHash, uint32 commitIndex) internal {
+        Commit[] storage cc = commitLocker[commitHash];
+        require(cc.length > 0, "Invalid commit hash");
```
ivan-homoliak-sutd, line 425: This and the following line will never evaluate to false since the function verifyReveal ensures that returned index is always within the allowed range. The same for timestamp check. Since require is related to the workflow of the code and asserts check internal errors, I'd recommend here to use asserts instead of requires. So, it will not confuse the auditors seeking for the code logic.

code/contracts/ONEWallet.sol
```diff
@@ -351,25 +393,40 @@ contract ONEWallet is TokenTracker {
...
+        require(cc.length > commitIndex, "Invalid commitIndex");
+        Commit storage c = cc[commitIndex];
+        require(c.timestamp > 0, "Invalid commit timestamp");
+        // should not happen
         uint32 index = uint32(c.timestamp) / interval - t0;
         _incrementNonce(index);
         _cleanupNonces();
```
ivan-homoliak-sutd, line 432: In function _cleanupNonces (4th line), can it happen that (indexMinUnadjusted > t0) is not true?

code/contracts/ONEWallet.sol
```diff
@@ -351,25 +393,40 @@ contract ONEWallet is TokenTracker {
...
+            bytes32 expectedVerificationHash = keccak256(bytes.concat(c.paramsHash, eotp));
+            if (c.verificationHash != expectedVerificationHash) {
+                // Invalid entry. Ignore
+                continue;
+            }
+            require(c.paramsHash == paramsHash, "Parameter hash mismatch");
+            uint32 counter = c.timestamp / interval - t0;
+            require(counter == index, "Index - timestamp mismatch");
+            uint8 expectedNonce = nonces[counter];
+            require(nonce >= expectedNonce, "Nonce too low");
```
ivan-homoliak-sutd, line 414: Why do you enforce this condition? It might have negative consequences on the use case of the shared wallet that you mentioned before. I'd say that since each nonce corresponds to its own EOTP there is no need for such a constraint. EOTPs aggregated by the Merkle enable selective acknowledgment (see SmartOTPs paper.)

code/contracts/ONEWallet.sol
```diff
@@ -117,35 +118,66 @@ contract ONEWallet is TokenTracker {
...
+        bytes32[] memory paramHashes = new bytes32[](numCommits);
+        bytes32[] memory verificationHashes = new bytes32[](numCommits);
+        uint32[] memory timestamps = new uint32[](numCommits);
+        bool[] memory completed = new bool[](numCommits);
+        uint32 index = 0;
+        for (uint32 i = 0; i < commits.length; i++) {
+            Commit[] storage cc = commitLocker[commits[i]];
+            for (uint32 j = 0; j < cc.length; j++) {
+                Commit storage c = cc[j];
+                hashes[index] = c.hash;
```
ivan-homoliak-sutd, line 142: Hash can be removed from the Commit structure. It is never utilized for any verification, so storage can be saved. Moreover, the hash is a key to commitLocker map, so you already have that info.

code/contracts/ONEWallet.sol
```diff
@@ -117,35 +118,66 @@ contract ONEWallet is TokenTracker {
...
+                paramHashes[index] = c.paramsHash;
```
ivan-homoliak-sutd, line 143: Commit structure contains paramHash, which should hide the details of the operation and thus ensure privacy. However, this is not met in all the cases, since paramHash might contain a repeated transfer to the same address with the same amount or the operation with a single parameter (recovery), which can be easily determined by the attacker. It can be resolved by adding some random 32B salt into paramHash, generated by the client.

Anyway, I still think that there is no reason for hiding the operation details in commit:
1) confirmation by OTPs is supposed to be used only for small amounts, so the attacker will not pay for overturning the blockchain to make some double-spending.
2) if you compare it with the native transfer of ONE, signed by ECDSA, the details of the transfer are leaked as well, so the MITM attacker can eclipse the victim in the same manner as in the case of OTPs.

ivan-homoliak-sutd (commented):
Security-wise, I do not see any critical issues, so I provide suggestions for optimizations and questions to possible minor issues.

code/contracts/ONEWallet.sol
```diff
@@ -117,35 +118,66 @@ contract ONEWallet is TokenTracker {
...
-    function commit(bytes32 hash, bytes32 paramsHash) external {
+    function commit(bytes32 hash, bytes32 paramsHash, bytes32 verificationHash) external {
         _cleanupCommits();
-        Commit storage c = commitLocker[hash];
-        require(c.timestamp == 0 && !c.completed, "Commit already exists");
-        Commit memory nc = Commit(hash, paramsHash, uint32(block.timestamp), false);
+        Commit memory nc = Commit(hash, paramsHash, verificationHash, uint32(block.timestamp), false);
         require(commits.length < MAX_COMMIT_SIZE, "Too many commits");
         commits.push(hash);
-        commitLocker[hash] = nc;
```
ivan-homoliak-sutd, line 148: The hash under you store to commitLocker is sufficient when it contains only h(EOTP). The true commitment hash that commits to the parameters using EOTP is stored as the verification hash in commit structure. So, removing the neighbor and indexWithNonce from hash should be made. (Also, the hash field can be dropped from the commit structure - as I already mentioned in other review.)

code/contracts/ONEWallet.sol
polymorpher: Discussed over chat and I agree.

code/contracts/ONEWallet.sol
polymorpher: Also discussed over chat, and I agree.

code/contracts/ONEWallet.sol
polymorpher: Yes, this is a remnant of an early version to prevent some types of attacks. It should be modified to `require(nonce < maxOperationsPerInterval, "Nonce is invalid");`

code/contracts/ONEWallet.sol
polymorpher: I agree

code/contracts/ONEWallet.sol
polymorpher: If it is not true, `indexMin` would remain at 0. `_cleanupNonces` would do nothing, so there is no harm.

code/contracts/ONEWallet.sol
polymorpher: I agree in parts. A salt seems reasonable. For the later part regarding reasons to hide operations at commit time, note a distinction between the situation here and that of ECDSA transfer in a single transaction:

1. There is one block confirmation difference (ECDSA transaction would be executed in the same block, here the reveal only happens in the next block).
2. Gas is already paid for the commit (which will be deducted from the wallet, as soon as Account Abstraction is implemented) prior to the occurrence of the fund-transfer transaction. 

These two factors made it very easy for an opponent (in an application which the user is competing with others) to act in response to the user's intent (consider auctions, gamblings, games, etc.). Previously (or in the ECDSA case) the opponent has to operate a very active mining (validation) node, actively scan and parse all transactions in the current block, and pay a much higher gas if the attacker wants to utilize the revealed information to do anything (such as changing price for an auction, place a slightly higher bid, changing odds in a game the attacker hosts, or make advanced movements in a game, etc.). Now, with the parameters revealed in commit, the attacker can just lay back and examine a block that is already confirmed and index where the commit resides.

code/contracts/ONEWallet.sol
polymorpher: On a second thought - I think indexWithNonce should be kept. There may be repetitions in EOTP (especially if double OTP and controlled randomness are not used), though the probability that they occur less than two time intervals apart are very unlikely. The computation for 32 bytes more data in keccak256 is cheap anyway. And for backward compatibility reasons, it doesn't hurt to have 32 more bytes for neighbor as well. Let's keep this as is for now.

code/contracts/ONEWallet.sol
polymorpher: My previous comment in this thread is wrong. It would always be true that `nonce < maxOperationsPerInterval ` since nonce is the remainder of modulus `maxOperationsPerInterval`.  I think the only remaining purpose now for nonce is to track how many operations are already performed in the current time interval. The difference this line makes is whether we enforce nonce to be monotonically increasing. Suppose if we remove this line, then clients (possessing proofs) would be free to choose which nonce to use. It would be chaotic, since clients would have no idea what nonces other clients are using. If two clients choose the same nonce at the same time, one of them would fail. So in that situation, if two clients are trying nonces incrementally, it would not really be any better than the current situation enforced by this line: two clients would be forced to query the current nonce of the current block, and one of them would be forced to wait until the other finishes. On the other hand, if two clients coordinate with each other and try different nonces at the same time, then both transactions could go through. But then the client implementation would get complicated - WebSocket or WebRTC would have to be used for coordination, and we would inevitably need a server to broker the communication before the clients connect with each other. In view of that, I don't think we should remove the current constraint imposed by this line, until we do have proper WebSocket or WebRTC implementations to coordinate multiple clients. For now, the user(s) should coordinate manually and let one transaction finishes before the other in the multi-client use case. This approach is better for security reason as well - an attacker with the capability to intercept OTPs and read client side data would be forced to block the commit from the user's client, as otherwise the user's transaction would race with the attacker's. It makers this kind of attack, which is already very difficult to pull off, even easier to detect and harder to succeed.

polymorpher:
All comments are addressed in branch https://github.com/polymorpher/one-wallet/tree/v9.0 , except those commented above

PR #65, opened 2021-08-04
Add use two otp options on create wallet

haolinj:
This is working in progress as part of client-security branch.


code/client/src/pages/Create.jsx
```diff
@@ -60,52 +76,72 @@ const Create = () => {
...
 
   const [deploying, setDeploying] = useState()
 
   const otpRef = useRef()
 
-  const getQRCodeUri = () => {
+  const getQRCodeUri = (otpSeed) => {
     // otpauth://TYPE/LABEL?PARAMETERS
-    return `otpauth://totp/${name}?secret=${b32.encode(seed)}&issuer=Harmony`
+    return `otpauth://totp/${name}?secret=${b32.encode(otpSeed)}&issuer=Harmony`
```
polymorpher, line 92: Need to use a slightly different name here. e.g. instead of `${name}`, use `${name} (2nd)`

code/client/src/pages/Create.jsx
```diff
@@ -60,52 +76,72 @@ const Create = () => {
...
     if (code.padStart(6, '0') !== otp.padStart(6, '0')) {
-      console.log(`Expected: ${code}. Got: ${otp}`)
       message.error('Code is incorrect. Please try again.')
-      setOtp('')
+
+      otpRef?.current?.focusInput(0)
+    } else if (doubleOtp && !settingUpSecondOtp) {
+      setSection(sectionViews.setupSecondOtp)
+
+      setName(generateNewOtpName())
```
polymorpher, line 140: should not generate a new name. The name is the (local) name of the wallet to help the user identify the wallet. It should be the same for both OTP entries on the authenticator, and the same name should be stored in redux state

code/client/src/pages/Create.jsx
haolinj: So the `name` is the same, but append `(2nd)`?

code/client/src/pages/Create.jsx
```diff
@@ -232,6 +272,34 @@ const Create = () => {
             </Row>
           </Space>
         </Row>
+        <Row justify='center'>
+          <Checkbox onChange={() => setDoubleOtp(!doubleOtp)}>
+            <Hint>Use two One Time Passwords for enhanced security</Hint>
```
polymorpher, line 277: Should let user know what to expect - i.e. they will need to scan again on the next page. Otherwise they tick the box and nothing happens

code/client/src/pages/Create.jsx
polymorpher: yes

code/client/src/pages/Create.jsx
haolinj: Do you mean the `name` is the same, but the second one has `(2nd)` appended?

polymorpher: yes

code/client/src/pages/Create.jsx
```diff
@@ -60,52 +76,72 @@ const Create = () => {
...
     if (code.padStart(6, '0') !== otp.padStart(6, '0')) {
-      console.log(`Expected: ${code}. Got: ${otp}`)
       message.error('Code is incorrect. Please try again.')
-      setOtp('')
+
+      otpRef?.current?.focusInput(0)
+    } else if (doubleOtp && !settingUpSecondOtp) {
+      setSection(sectionViews.setupSecondOtp)
+
+      setName(`${name} (2nd)`)
```
polymorpher, line 140: It would be incorrect to change the name state variable because the wallet would be saved as `${name} (2nd)` in redux

haolinj: Good point. So we still want to save the name as original `name` in the state, but the authenticator would have a different name for the second OTP (that is only append (2nd) text).

code/client/src/pages/Show.jsx
```diff
@@ -134,10 +136,13 @@ const Show = () => {
       setInputAmount(formatted)
     }
   }
+
   const resetOtp = () => {
     setOtpInput('')
+    setOtp2Input('')
```
polymorpher, line 142: otp2Ref is missing

polymorpher: you are right

polymorpher: we should auto focus to second box, after first box has 6 digits

polymorpher:
Everything works - very good job

PR #66, opened 2021-08-05
Client security: enable randomness by default 

polymorpher:


polymorpher:
Recover randomness is done at worker - overall experience is quite smooth, given the current parameters (18 bits randomness for SHA256, 16 bits for argon2). By default wallets are created with SHA256 hasher. We can move to argon2 later after its runtime is accelerated by another 5-10x - see my notes in #64 for how

PR #77, opened 2021-08-12
Issue/75 less ambiguous wallet names

haolinj:


polymorpher:
Looks good, except the address should show the full address as the tooltip when hovered over. I suggest creating a new component for Address, since another change in this version requires the address to be displayed in ETH form when hovered over for more than 1 second.

Actually, let's replace the explorer link with a toggle to the Address component. Toggle it when the user clicks the address, instead of sending them to the explorer. When the toggle is on, show more buttons to the right: 

1. copy (already here)
2. go to explorer in new window (use the address consistent with the current style)
3. switch address style (toggle button) between bech32 (one1...) and hex (0x...)

use some sensible icon from antdesign icon collections

haolinj:
> Looks good, except the address should show the full address as the tooltip when hovered over. I suggest creating a new component for Address, since another change in this version requires the address to be displayed in ETH form when hovered over for more than 1 second.
> 
> Actually, let's replace the explorer link with a toggle to the Address component. Toggle it when the user clicks the address, instead of sending them to the explorer. When the toggle is on, show more buttons to the right:
> 
>     1. copy (already here)
> 
>     2. go to explorer in new window (use the address consistent with the current style)
> 
>     3. switch address style (toggle button) between bech32 (one1...) and hex (0x...)
> 
> 
> use some sensible icon from antdesign icon collections

Yep, makes sense. I've updated the PR. Please re-review when you have time.

polymorpher:
Looks good. Let's use this everywhere where addresses are displayed.

PR #79, opened 2021-08-15
Issue/75 address input for known address

haolinj:


code/client/src/components/AddressInput.jsx
```diff
@@ -0,0 +1,103 @@
...
+  const existingKnownAddress = useCallback((oneAddress) => {
+    return addressOptions.find((wallet) =>
+      util.safeOneAddress(wallet.address) === oneAddress || wallet.knownOneAddress === oneAddress)
+  }, [addressOptions])
+
+  const onSelectAddress = useCallback((oneAddress) => {
+    const validOneAddress = util.safeOneAddress(oneAddress)
+
+    if (validOneAddress && !existingKnownAddress(validOneAddress)) {
+      dispatch(walletActions.addKnownAddress({ knownAddressKey, oneAddress: validOneAddress }))
```
polymorpher, line 55: We should keep all addresses stored in redux state in normalized format (0x....). Convert to bech32 (one1...) only during display.

code/client/src/components/AddressInput.jsx
```diff
@@ -0,0 +1,103 @@
...
+      onSelect={onSelectAddress}
+    >
+      {
+        addressOptions
+          .filter((wallet) => wallet.network === network && notCurrentWallet(wallet.knownOneAddress || wallet.address))
+          .map((wallet, index) => {
+            const addr = wallet.knownOneAddress || util.safeOneAddress(wallet.address)
+
+            return (
+              <Select.Option key={index} value={wallet.knownOneAddress || util.safeOneAddress(wallet.address)}>
```
polymorpher, line 85: Can we also allow removing an address? e.g. adding a `x` at the end. Let's do it later in a separate PR. It would be nice to be able to edit address label as well.

code/client/src/constants/wallet.js
```diff
@@ -7,5 +7,9 @@ export default {
   maxTransferAttempts: 3,
   checkCommitInterval: 5000,
   fetchBalanceFrequency: 15000,
-  fetchDelaysAfterTransfer: [0, 2000, 5000]
+  fetchDelaysAfterTransfer: [0, 2000, 5000],
+  knownAddressKeys: {
```
polymorpher, line 11: It is not necessary to separate them. In reality, I find I do a lot of transfer between recovery address and 1wallet addresses. Let's drop the keys and use a single array to store them.

code/client/src/state/modules/wallet/reducers.js
```diff
@@ -146,6 +148,17 @@ const reducer = handleActions(
       ...state,
       provider: action.payload,
     }),
+
+    [walletActions.addKnownAddress]: (state, action) => ({
+      ...state,
+      knownAddresses: {
+        ...state.knownAddresses || {},
+        [action.payload.knownAddressKey]: [
```
polymorpher, line 156: see above comments on single object, and array of keys v.s. object of [address => {...}]

code/client/src/components/AddressInput.jsx
polymorpher, line 55: I think it would be beneficial to store the addresses as key-value format, with key equal to the normalized address, and value equal to an object consisting of: label (e.g. wallet name), network (i.e. currently selected network), creationTime (i.e. when it is added), lastUsedTime, and numUsed (e.g. increment by one each time it is selected). The reason is that we will need to allow users to re-label the address later, and later they would expect us to intelligently rank the addresses based on how frequently they use each address, or how old they are. There might be more extensions in the future. Storing them as objects would allow us easily add more extensions, whereas array would not.

polymorpher: network is quite important - it must be stored, to allow the options automatically filter by network. We don't want the users accidentally send to an address on a different network - the money would have been lost

code/client/src/components/AddressInput.jsx
```diff
@@ -0,0 +1,103 @@
...
+  const notCurrentWallet = useCallback((inputWalletAddress) =>
+    !currentWallet || currentWallet?.address !== inputWalletAddress,
+  [currentWallet])
+
+  /**
+   * This is interim wallet object that uses known wallet addresses to be used to render the address selection with
+   * existing wallet object. The address key is unique as knownOneAddress. Note that this will not be saved as actual wallet.
+   */
+  const knownWallets = (knownAddresses ? knownAddresses[knownAddressKey] || [] : []).map((knownOneAddress) => ({
+    network,
```
polymorpher, line 40: see below comments on network - the network should be retrieved from state, based on which network was selected when the address was added. The network should not be just added based on the current selected network right now

code/client/src/components/AddressInput.jsx
polymorpher: let's at  least support deletion - otherwise the list of addresses could add up and become overwhelming quite quickly

code/client/src/components/AddressInput.jsx
```diff
@@ -0,0 +1,103 @@
...
+import { Select } from 'antd'
+import React, { useCallback } from 'react'
+import { useDispatch, useSelector } from 'react-redux'
+import walletActions from '../state/modules/wallet/actions'
+import util, { useWindowDimensions } from '../util'
+
+/**
+ * Renders address input that provides type ahead search for any known ONE addresses.
+ * Known addresses are addresses that have been entered by user for at least once.
+ * If an address is entered in none one1... format, the address will be converted and saved as one1... format.
```
polymorpher, line 11: see below comment on always using hex format in state

code/client/src/components/WalletAddress.jsx
```diff
@@ -51,25 +52,25 @@ const WalletAddressOptions = ({ displayAddress, address, network, onAddressStyle
...
 
 const MOUSE_HOVER_DETECTION_DELAY = 1000
+
 /**
  * Renders the provided wallet's address in either ONE style or normal.
  * Provides the ability to copy the address and link to wallet explorer.
  */
-const WalletAddress = ({ isMobile, wallet, network }) => {
-  const oneAddress = util.safeOneAddress(wallet.address)
+const WalletAddress = ({ isMobile, wallet, network, addressOverride, shortAddress }) => {
```
polymorpher, line 60: It seems only two things from wallet are used here: address, name. And the name is solely used to determine whether an address should be shortened. Wouldn't it be cleaner if only `address` and `shorten` are arguments?

I don't think `network` is necessary either. You can use a selector inside this component  to get the current network

code/client/src/components/AddressInput.jsx
haolinj: Done.

code/client/src/components/AddressInput.jsx
haolinj: Added the network to known address object and used as filter when display them.

code/client/src/components/AddressInput.jsx
haolinj: Changed to use the object to store addresses.

haolinj: Address changed to object in state.

code/client/src/components/AddressInput.jsx
haolinj: Adding a `x` button at the end in the Select option does not work at the moment. When you click on the delete button it also acts as select, which will add the deleted address back. I haven't found a way to work around this yet.

code/client/src/components/WalletAddress.jsx
haolinj: Good point, I've refactored this component.

code/client/src/constants/wallet.js
haolinj: Removed this key separation.

code/client/src/state/modules/wallet/reducers.js
haolinj: Changed to store known address as object with network and usage details.

code/client/src/state/modules/wallet/reducers.js
```diff
@@ -149,16 +149,31 @@ const reducer = handleActions(
...
+    [walletActions.setKnownAddress]: (state, action) => ({
       ...state,
       knownAddresses: {
-        ...state.knownAddresses || {},
-        [action.payload.knownAddressKey]: [
-          ...state.knownAddresses?.[action.payload.knownAddressKey] || [],
-          action.payload.oneAddress
-        ]
+        ...state.knownAddresses,
+        [action.payload.address]: {
```
polymorpher, line 156: guard for undefined `payload.address`?

haolinj: The action payload should be filtered before it reaches the reducer. Trying to keep reducer logic free.

code/client/src/pages/Show.jsx
```diff
@@ -571,9 +570,8 @@ const Show = () => {
             <Label><Hint>To</Hint></Label>
             <AddressInput
               addressValue={transferTo}
-              setAddressCallback={(value) => setTransferTo(value)}
+              setAddressCallback={setTransferTo}
```
polymorpher, line 573: Would that cause a warning "modifying another component's state inside a component"?

haolinj: I haven't seen this warning. But it is a call back to set the value in this component, and this component's value is being supplied as prop to the `AddressInput` component.

polymorpher:
Looks good, but I am getting

```
react-dom.development.js:67 Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.
```

haolinj:
> Looks good, but I am getting
> 
> `react-dom.development.js:67 Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.`

I will double check and fix it.

code/client/src/components/AddressInput.jsx
```diff
@@ -0,0 +1,180 @@
...
+            const addr = util.safeOneAddress(knownAddress.address)
+            const longAddressLabel = knownAddress.label ? `(${knownAddress.label}) ${addr}` : addr
+            const shortenAddressLabel = knownAddress.label ? `(${knownAddress.label}) ${util.ellipsisAddress(addr)}` : util.ellipsisAddress(addr)
+            const displayLabel = util.shouldShortenAddress({ walletName: knownAddress.label, isMobile })
+              ? shortenAddressLabel
+              : longAddressLabel
+
+            return (
+              <Select.Option key={index} value={util.safeOneAddress(knownAddress.address)}>
+                <Space size='small' align='baseline'>
```
polymorpher, line 126: `justify='spaceBetween'`? Looks weird when checkmarks are unaligned

haolinj: Good point, I did not what to set there.

polymorpher: actually it will need a `Row` and it is `space-between`

haolinj: Yeah, I thought it would be `Grid` layout. I will play around it and try to make it align.

polymorpher:
> > Looks good, but I am getting
> > `react-dom.development.js:67 Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.`
> 
> I will double check and fix it.

It is reporting `setAddressCallback` is changing on every render. Since it is a state-setter, and the function is quite large, it might not be a good idea to have that in the dependency array 

```
function dispatchAction(fiber, queue, action) {
  {
    if (typeof arguments[3] === 'function') {
      error("State updates from the useState() and useReducer() Hooks don't support the " + 'second callback argument. To execute a side effect after ' + 'rendering, declare it in the component body with useEffect().');
    }
  }

  var eventTime = requestEventTime();
  var lane = requestUpdateLane(fiber);
  var update = {
    lane: lane,
    action: action,
    eagerReducer: null,
    eagerState: null,
    next: null
  }; // Append the update to the end of the list.

  var pending = queue.pending;

  if (pending === null) {
    // This is the first update. Create a circular list.
    update.next = update;
  } else {
    update.next = pending.next;
    pending.next = update;
  }

  queue.pending = update;
  var alternate = fiber.alternate;

  if (fiber === currentlyRenderingFiber$1 || alternate !== null && alternate === currentlyRenderingFiber$1) {
    // This is a render phase update. Stash it in a lazily-created map of
    // queue -> linked list of updates. After this render pass, we'll restart
    // and apply the stashed updates on top of the work-in-progress hook.
    didScheduleRenderPhaseUpdateDuringThisPass = didScheduleRenderPhaseUpdate = true;
  } else {
    if (fiber.lanes === NoLanes && (alternate === null || alternate.lanes === NoLanes)) {
      // The queue is currently empty, which means we can eagerly compute the
      // next state before entering the render phase. If the new state is the
      // same as the current state, we may be able to bail out entirely.
      var lastRenderedReducer = queue.lastRenderedReducer;

      if (lastRenderedReducer !== null) {
        var prevDispatcher;

        {
          prevDispatcher = ReactCurrentDispatcher$1.current;
          ReactCurrentDispatcher$1.current = InvalidNestedHooksDispatcherOnUpdateInDEV;
        }

        try {
          var currentState = queue.lastRenderedState;
          var eagerState = lastRenderedReducer(currentState, action); // Stash the eagerly computed state, and the reducer used to compute
          // it, on the update object. If the reducer hasn't changed by the
          // time we enter the render phase, then the eager state can be used
          // without calling the reducer again.

          update.eagerReducer = lastRenderedReducer;
          update.eagerState = eagerState;

          if (objectIs(eagerState, currentState)) {
            // Fast path. We can bail out without scheduling React to re-render.
            // It's still possible that we'll need to rebase this update later,
            // if the component re-renders for a different reason and by that
            // time the reducer has changed.
            return;
          }
        } catch (error) {// Suppress the error. It will throw again in the render phase.
        } finally {
          {
            ReactCurrentDispatcher$1.current = prevDispatcher;
          }
        }
      }
    }

    {
      // $FlowExpectedError - jest isn't a global, and isn't recognized outside of tests
      if ('undefined' !== typeof jest) {
        warnIfNotScopedWithMatchingAct(fiber);
        warnIfNotCurrentlyActingUpdatesInDev(fiber);
      }
    }

    scheduleUpdateOnFiber(fiber, lane, eventTime);
  }
}
```

code/client/src/components/AddressInput.jsx
```diff
@@ -0,0 +1,180 @@
...
+    walletsNotInKnownAddresses.forEach((wallet) => {
+      dispatch(walletActions.setKnownAddress({
+        label: wallet.name,
+        address: wallet.address,
+        network: wallet.network,
+        creationTime: wallet.effectiveTime,
+        numUsed: 0
+      }))
+    })
+  }, [knownAddresses, wallets, dispatch])
```
polymorpher, line 66: Turns out the issue is with this dependency `knownAddresses`. There might be some invalid wallets left in the user's state (e.g. my browser). If the invalid wallet has no valid `address`, it would become part of `walletsNotInKnownAddresses` and sent to dispatch for `setKnownAddress`. The reducer in `setKnownAddress` can't add that invalid wallet's address to known addresses, but would create a new object for `knownAddresses`, therefore trigger another `useEffect` here, hence causing an infinite loop

haolinj: I've made this effect run once when the component load.

code/client/src/components/AddressInput.jsx
haolinj: Replaced with Grid layout and removed the check button, replaced with text button that spans the whole Col space.

code/client/src/components/AddressInput.jsx
```diff
@@ -0,0 +1,179 @@
...
+                </Row>
+              </Select.Option>
+            )
+          })
+      }
+      {
+        showSelectManualInputAddress
+          ? (
+            <Select.Option key='address-value' value={addressValue.value}>
+              <Space size='small' align='baseline'>
```
polymorpher, line 162: onClick needs to be on Space, otherwise the function won't trigger

code/client/src/components/AddressInput.jsx
```diff
@@ -0,0 +1,179 @@
...
+            return (
+              <Select.Option key={index} value={util.safeOneAddress(knownAddress.address)}>
+                <Row gutter={16} align='left'>
+                  <Col span={!displayDeleteButton ? 24 : 21}>
+                    <Tooltip title={addr}>
+                      <Button
+                        block
+                        type='text'
+                        style={{ textAlign: 'left' }}
+                        onClick={() => onSelectAddress({ value: addr, label: longAddressLabel, key: index })}
```
polymorpher, line 136: same as below

code/client/src/components/AddressInput.jsx
```diff
@@ -0,0 +1,179 @@
...
+          })
+      }
+      {
+        showSelectManualInputAddress
+          ? (
+            <Select.Option key='address-value' value={addressValue.value}>
+              <Space size='small' align='baseline'>
+                {addressValue.value}
+                <Button type='text' onClick={() => onSelectAddress(addressValue)}>
+                  <CheckOutlined />
```
polymorpher, line 165: no longer needed

code/client/src/pages/Create.jsx
```diff
@@ -345,29 +346,22 @@ const Create = () => {
...
+            <AddressInput
+              addressValue={lastResortAddress}
+              setAddressCallback={setLastResortAddress}
+              extraSelectOptions={
+                [
+                  <Select.Option key='later' value=''>
+                    <Space size='small' align='baseline'>
+                      I want to do this later in my wallet
+                      <Button type='text' onClick={() => setLastResortAddress({ value: '', label: 'I want to do this later in my wallet' })}>
+                        <CheckOutlined />
```
polymorpher, line 358: no longer needed, but since we are removing this option altogether - you can update it in the next PR

code/client/src/pages/Create.jsx
```diff
@@ -345,29 +346,22 @@ const Create = () => {
...
-                return (
-                  <Select.Option key={k} value={util.safeOneAddress(wallets[k].address)}>
-                    ({wallets[k].name}) {isMobile ? util.ellipsisAddress(addr) : addr}
+            <AddressInput
+              addressValue={lastResortAddress}
+              setAddressCallback={setLastResortAddress}
+              extraSelectOptions={
+                [
+                  <Select.Option key='later' value=''>
+                    <Space size='small' align='baseline'>
```
polymorpher, line 355: same as above

code/client/src/pages/Create.jsx
haolinj: Yes, forgot to remove it, will remove and push.

polymorpher:
Looks good

PR #87, opened 2021-08-20
Big frontend refactor - break down wallet components to 11 subcomponents

polymorpher:


polymorpher:
Tested transfer, token operations, recovery, set recovery address, and auto-triggered recovery. Bugs are fixed. Fully functional.

Issue #76, opened 2021-08-12
Upcoming changes for v0.8.3: Human-readable domain names registration for 1wallet accounts


polymorpher:
After a wallet is created, the user may purchase a domain name powered by [crazy.one](https://crazy.one) in the wallet UI to associate the wallet with that name, such that other 1wallet users can transfer funds to the domain name instead of being forced to use the long address (`one1....` or `0x....`) for transferring funds. A domain name can be as cheap as 1 ONE. For more details, please visit https://medium.com/harmony-one/harmony-community-launches-crazy-one-the-first-subdomain-nft-7db95bc53326

A proposed flow is as the following:

Next to the wallet's address, a "purchase" button will be displayed next to the “copy” button. The purchase button is only displayed if the wallet is not associated with a domain name. The same button is also be visible under “About” tab of the wallet. 

When the user clicks the purchase button, a new screen (similar to transfer or setRecovery) is displayed to guide the user through the domain purchase step-by-step. First, the user needs to choose a name. The UI would look similar to `[input box].crazy.one`. The UI would give the user a quote immediately as the user types the name, and inform the user whether the domain is available or not. When the user finalizes the name, she can click the buy button, at which time the wallet will check whether the user has sufficient balance. If so, the wallet moves to the standard transfer screen with title “associate wallet with domain xxx”. The user needs to follow the standard procedure for transfer (i.e. requiring OTP input) to complete the purchase.

polymorpher:
Note: source code for crazy.one, which has examples of interaction flows with the ENS contracts, is available at https://github.com/harmony-one/crazy-one (made by @givp )

polymorpher:
Done

PR #89, opened 2021-08-22
Issue/76 lookup resolve domain addressinput

haolinj:
#76

### DIFF
--- a/code/client/src/components/AddressInput.jsx
+++ b/code/client/src/components/AddressInput.jsx
@@ -220,8 +220,8 @@ const AddressInput = ({ setAddressCallback, currentWallet, addressValue, extraSe
       : longAddressLabel
 
     return (
-      <Select.Option key={displayText} value={filterValue}>
-        <Row gutter={16} align='left'>
+      <Select.Option key={displayText} value={filterValue} style={{ padding: 0 }}>
+        <Row align='left'>
           <Col span={!displayDeleteButton ? 24 : 21}>
             <Tooltip title={oneAddress}>
               <Button
--- a/code/client/src/components/AddressInput.jsx
+++ b/code/client/src/components/AddressInput.jsx
@@ -1,10 +1,13 @@
 import { CloseOutlined, SearchOutlined } from '@ant-design/icons'
-import { Select, Button, Tooltip, Row, Col } from 'antd'
-import React, { useCallback, useEffect } from 'react'
+import { Select, Button, Tooltip, Row, Col, Spin, Typography } from 'antd'
+import React, { useCallback, useEffect, useState } from 'react'
 import { useDispatch, useSelector } from 'react-redux'
 import walletActions from '../state/modules/wallet/actions'
 import util, { useWindowDimensions } from '../util'
 import WalletConstants from '../constants/wallet'
+import api from '../api'
+
+const { Text } = Typography
 
 /**
  * Renders address input that provides type ahead search for any known addresses.
@@ -12,6 +15,8 @@ import WalletConstants from '../constants/wallet'
  */
 const AddressInput = ({ setAddressCallback, currentWallet, addressValue, extraSelectOptions }) => {
   const dispatch = useDispatch()
+
+  const [searchingAddress, setSearchingAddress] = useState(false)
 
   const wallets = useSelector(state => Object.keys(state.wallet.wallets).map((k) => state.wallet.wallets[k]))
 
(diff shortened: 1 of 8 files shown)


code/client/src/components/AddressInput.jsx
```diff
@@ -24,13 +34,53 @@ const AddressInput = ({ setAddressCallback, currentWallet, addressValue, extraSe
...
+            domainName,
+            filterValue: searchValue,
+            selected: false
+          })
+        }
+
+        if (!isEmpty(searchValue)) {
+          const resolvedAddress = await api.blockchain.domain.resolve({ name: searchValue })
+
+          if (resolvedAddress && resolvedAddress !== '0x0000000000000000000000000000000000000000') {
```
polymorpher, line 65: util.isEmptyAddress

code/client/src/components/AddressInput.jsx
```diff
@@ -24,13 +34,53 @@ const AddressInput = ({ setAddressCallback, currentWallet, addressValue, extraSe
...
+
+          setAddressCallback({
+            value: validAddress,
+            domainName,
+            filterValue: searchValue,
+            selected: false
+          })
+        }
+
+        if (!isEmpty(searchValue)) {
```
polymorpher, line 62: could do some preliminary filtering to reduce queries to blockchain, e.g. `searchValue.includes(`${ONEConstants.Domain.DEFAULT_PARENT_LABEL}.${ONEConstants.Domain.DEFAULT_TLD} )`

code/client/src/components/AddressInput.jsx
```diff
@@ -24,13 +34,53 @@ const AddressInput = ({ setAddressCallback, currentWallet, addressValue, extraSe
...
+              value: resolvedAddress,
+              domainName: searchValue,
+              filterValue: searchValue,
+              selected: false
+            })
+          }
+        }
+
+        setSearchingAddress(false)
+      } catch (e) {
```
polymorpher, line 76: let's log it to console (which should be also captured by Sentry) so we know if the resolver is temporarily down or if we are blocked due to 429

code/client/src/components/AddressInput.jsx
```diff
@@ -42,27 +92,48 @@ const AddressInput = ({ setAddressCallback, currentWallet, addressValue, extraSe
...
-    // Init the known address entries for existing wallets.
-    walletsNotInKnownAddresses.forEach((wallet) => {
-      dispatch(walletActions.setKnownAddress({
-        label: wallet.name,
-        address: wallet.address,
-        network: wallet.network,
-        creationTime: wallet.effectiveTime,
-        numUsed: 0
+      const knownAddressesWithoutDomain = existingKnownAddresses.filter((knownAddress) =>
+        !isEmpty(existingKnownAddresses.domain?.name)
```
polymorpher, line 105: typo?

haolinj: Yes, fixed.

code/client/src/components/AddressInput.jsx
```diff
@@ -42,27 +94,48 @@ const AddressInput = ({ setAddressCallback, currentWallet, addressValue, extraSe
...
 
-    // Init the known address entries for existing wallets.
-    walletsNotInKnownAddresses.forEach((wallet) => {
-      dispatch(walletActions.setKnownAddress({
-        label: wallet.name,
-        address: wallet.address,
-        network: wallet.network,
-        creationTime: wallet.effectiveTime,
-        numUsed: 0
+      const knownAddressesWithoutDomain = existingKnownAddresses.filter((knownAddress) =>
```
polymorpher, line 106: we should probably add a time condition later to reduce queries, but it is probably fine for now

haolinj: Yeah, I agree, right now this is just bootstrap existing address that does not have a domain. We can use lookupTime or last lookupTime to limit the query.

polymorpher:
Looks good. Doing some tests.

polymorpher:
`Enter` should be as good as select. Right now it clears selection

polymorpher:
If I type the domain name of a known wallet (e.g. another 1wallet of mine), it fails to resolve the domain correctly to that wallet.

image
 
image

polymorpher:
It should also automatically trim extra spaces

image

polymorpher:
Should add support to Restore as well

image

polymorpher:
The rest seems to work very well

haolinj:
> If I type the domain name of a known wallet (e.g. another 1wallet of mine), it fails to resolve the domain correctly to that wallet.
> 
> image
> 
> image

This is due to no trim to the input address value, should be fixed.

polymorpher:
Restore: Showing technical errors on mainnet if address is not contract - the error should be hidden (should not use the handler in safeExec); Not recognizing address on Ganache (because of exception in api call)

image


image

polymorpher:
Looks good

Issue #78, opened 2021-08-12
Upcoming changes v0.9.0: upgradable wallet, domain name binding, conditionally triggered auto-recovery, recover tokens


polymorpher:
- [x] Upgradable wallet
- [x] Domain name binding
- [x] Conditionally triggered auto-recovery
- [x] Recover tokens

1. Wallets created on or after v0.9.0 will have the following functionality: Instead of creating new wallet and migrating assets, the user will be able to upgrade their wallets by clicking a "Upgrade" button when new versions are available. After the upgrade, the user would receive a new wallet address. The old address will forward all funds received to the new address. The old address will also automatically forward all ERC721 and 1155 tokens received, but not ERC20. However, the new address will be able to `pull` the ERC20 tokens from the old address (with the press of a button), by causing the old address to transfer those tokens to the new address. Each wallet will keep a record of which wallet will be forwarding funds to this wallet, and whether this wallet is forwarding funds to another wallet (upon an upgrade or recovery event)
2. Domain name binding: each wallet will keep a record of the domain name purchased and bound to this wallet, if any. This information will be displayed in "about" section of the wallet, and is retrievable through a public function on the smart contract.
3. Conditionally triggered auto-recovery: Sending 1.0 ONE from recovery address to the wallet is only effective when the wallet has not been performing any operation for a period of 14 days. "Performing any operation" is defined as successfully completing a reveal function call without being reverted. This is to prevent abuse of the recovery functionality, so that it is more likely than not that the wallet owner truly has lost access to the wallet, otherwise the security of the wallet would be entirely delegated to the recovery address, as discussed here: https://github.com/polymorpher/one-wallet/pull/56#discussion_r676150190 . This also ensures 1wallet treasury cannot tamper with users' funds if they actively use their wallets, since 1wallet treasury will be the default recovery address after v0.8.2 (https://github.com/polymorpher/one-wallet/issues/75)
4.  Recover tokens: During recovery, ERC721 and 1155 tokens will be automatically recovered to the recovery address. Automatically tracked ERC20 tokens will also be automatically recovered, but those untracked ERC20 tokens will not be. ERC20 tokens are automatically tracked when the user performs at least one transfer of the ERC20 token from the wallet. If the user received the token, but have never transferred it out, the token will not be automatically tracked. However, the recovery address can call a function on the wallet (after recovery is already done) to manually `pull` the tokens to the recovery address. If the recovery address is another 1wallet, that can be done by simply pressing a button on the UI.

[...]

PR #90, opened 2021-08-22
Upgradable wallets: frontend implementation and contract improvements (wallet graph); Fix dependency security issues; Robust token recovery & security improvement; Many more wallet operations

polymorpher:
Addresses: https://github.com/polymorpher/one-wallet/issues/78


polymorpher:
Looks good (self-reviewed). Several issues shall be addressed in the next PR:

1. domain transfer and renewal functions need to exist in the minimum upgradable version of the contract, otherwise domains can't be transferred to forwarded contracts later.
2. several external function calls need to have try-catch guard to prevent revert after reveal
3. backlink management code should be spun out in libraries

PR #91, opened 2021-08-22
Domain transfer and renewal; Safeguards around external calls; Reclaim domain operations; WalletGraph library; OTP recover randomness worker reset bug fix

polymorpher:
Addresses issues in https://github.com/polymorpher/one-wallet/pull/90#issuecomment-903262871 and to wrap up https://github.com/polymorpher/one-wallet/issues/78 and https://github.com/polymorpher/one-wallet/issues/76


polymorpher:
Looks good. Both @haolinj and I tested extensively. Domain transfer works as well, which means we can have upgraded wallets to claim domains previously purchased from linked addresses.

PR #99, opened 2021-08-29
SushiSwap UI (and base branch for other parts)

haolinj:
No actual swapping functionality added yet.


code/client/src/pages/Show/Swap.jsx
```diff
@@ -0,0 +1,268 @@
...
+ */
+const getSelectedTokenComputedBalance = (selectedToken, tokenBalances, oneWalletBalance) => {
+  try {
+    if (selectedToken && selectedToken.symbol === HarmonyONE.symbol) {
+      const computedBalance = util.computeBalance(oneWalletBalance, 0)
+
+      return computedBalance
+    }
+
+    const computedBalance = util.computeBalance(tokenBalances[selectedToken.address], selectedToken.decimals)
```
polymorpher, line 46: second parameter is price

code/client/src/pages/Show.jsx
```diff
@@ -24,8 +24,16 @@ import WalletTitle from '../components/WalletTitle'
...
 import Upgrade from './Show/Upgrade'
 import TransferDomain from './Show/TransferDomain'
+import Swap from './Show/Swap'
+
+const tabList = [
+  { key: 'coins', tab: 'Coins' },
+  { key: 'nft', tab: 'Collectibles' },
+  { key: 'about', tab: 'About' },
+  { key: 'help', tab: 'Recover' },
+  { key: 'swap', tab: 'Swap Coins' }
```
polymorpher, line 34: let's just call it Swap

code/client/src/pages/Show/Swap.jsx
```diff
@@ -0,0 +1,268 @@
...
+import { trim } from 'lodash'
+const { Text } = Typography
+
+// TODO: some token's images are not available, we may want a CDN or other service that can retrieve icons dynamically.
+const cryptoIconUrl = (symbol) => `https://qokka-public.s3-us-west-1.amazonaws.com/crypto-logos/${symbol.toLowerCase()}.png`
+
+const TokenLabel = ({ token }) => (
+  <>
+    <Image
+      style={{ display: 'inline', paddingRight: '5px' }}
```
polymorpher, line 20: image is distorted because of right padding

code/client/src/pages/Show/Swap.jsx
```diff
@@ -0,0 +1,268 @@
...
+      preview={false}
+      width={20}
```
polymorpher, line 22: 24 or 32

code/client/src/pages/Show/Swap.jsx
```diff
@@ -0,0 +1,268 @@
...
+/**
+ * Gets the supplied token balance. If the token is Harmony ONE, use the supplied wallet balance.
+ * Otherwise use token balances to compute the balance.
+ * @param {*} selectedToken selected token or wallet.
+ * @param {*} tokenBalances all tracked token balances.
+ * @param {*} oneWalletBalance Harmony ONE wallet balance in raw format.
+ * @returns computed balance.
+ */
+const getSelectedTokenComputedBalance = (selectedToken, tokenBalances, oneWalletBalance) => {
+  try {
```
polymorpher, line 39: There is no need to handle HarmonyONE specially. It has standard decimal (omit decimal parameter). Just pass in HarmonyONE as selectedToken

code/client/src/pages/Show/Swap.jsx
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    loadTrackedTokensMetadata()
+  }, [currentTrackedTokens, setCurrentTrackedTokens])
+
+  useEffect(() => {
+    const tokenBalance = getSelectedTokenComputedBalance(selectedTokenSwapFrom, tokenBalances, balance)
+
+    if (!tokenBalance) {
+      return '0'
```
polymorpher, line 148: Extraneous return value

code/client/src/pages/Show/Swap.jsx
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    setSelectedTokenBalance(tokenBalance.formatted)
```
polymorpher, line 151: Should set raw amount and display formatted. Transactions would rely on raw amounts

code/client/src/pages/Show/Swap.jsx
```diff
@@ -0,0 +1,268 @@
...
+      setSwapAmount(undefined)
+    } else {
+      setSwapAmount(amount)
+    }
+  }
+
+  const confirmSwap = useCallback(() => {
+    const swapAmountBalance = util.toBalance(swapAmount)
+    const computedTokenBalance = getSelectedTokenComputedBalance(selectedTokenSwapFrom, tokenBalances, balance)
+    const tokenBalance = util.toBalance(computedTokenBalance.formatted)
```
polymorpher, line 195: balance equals tokenBalance.balance. There is no need converting back and forth only to produce the same result

code/client/src/pages/Show/Swap.jsx
haolinj: `selectedTokenBalance` is used in input `value` for display, so it is a formatted value. When submit, the value is converted to raw amount.

code/client/src/pages/Show/Swap.jsx
```diff
@@ -0,0 +1,268 @@
...
+            showSearch
+            labelInValue
+            style={{ width: 280 }}
+            value={selectedTokenSwapTo}
+            onSearch={handleSearchSupportedTokens}
+          >
+            {swapOptions(supportedTokens, onSelectTokenSwapTo)}
+          </Select>
+        </Col>
+      </TallRow>
```
polymorpher, line 251: Need another column displaying the target amount

polymorpher:
The UI needs an exchange rate displayed when a target token is selected.

Can the UI be made more resembling SushiSwap? Display only symbol unless the dropdown boxes are expanded. Borderless dropdown box. 

Bigger icon, shorter column span for token. Bigger font and wider column span for input amount.

image

code/client/src/pages/Show/Swap.jsx
haolinj: `tokenBalances` does not seem to include Harmony's balance?

polymorpher: okay

code/client/src/pages/Show/Swap.jsx
haolinj: `computedTokenBalance` returns formatted balance, but we want the BN format for comparison?

haolinj:
> The UI needs an exchange rate displayed when a target token is selected.
> 
> Can the UI be made more resembling SushiSwap? Display only symbol unless the dropdown boxes are expanded. Borderless dropdown box.
> 
> Bigger icon, shorter column span for token. Bigger font and wider column span for input amount.
> 
> image

The UI is made with Antd and current theme of the wallet. The Sushi Swap is not really a dropdown, it opens a modal for token selection. Unless we want to add a modal.

code/client/src/pages/Show/Swap.jsx
haolinj: Target amount is based on the exchange rate. Do we have an API for that yet?

code/client/src/pages/Show/Swap.jsx
polymorpher: util.computeBalance returns the balance you passed in the parameters as the `balance` in its output. So if you only use `balance` field from `util.computeBalance`, the function is not doing anything. It is not guaranteed to be a BN. Also, `balance` everywhere is always referring to the raw, non-divisible amount. For ONE or ETH, it would be in wei.

code/client/src/pages/Show/Swap.jsx
polymorpher: Not yet, just mock it

code/client/src/pages/Show/Swap.jsx
haolinj: Good catch, will fix it.

code/client/src/pages/Show/Swap.jsx
```diff
@@ -52,27 +130,60 @@ const getSelectedTokenComputedBalance = (selectedToken, tokenBalances, oneWallet
...
-const optionButtonStyle = {
-  textAlign: 'left',
-  height: '50px'
-}
+/**
+ * Handles swap token selection. Either a "swap from" token or a "swap to" token can be handled.
+ * TODO: actual exchange rate should be fetched and updated here.
+ * If the selected token changes, the target swap amount will be automatically updated based on fetched exchange rate.
+ */
+const handleSwapTokenSelected = ({
```
polymorpher, line 163: If this function and the one above depend on so many state variables and functions, it might make more sense to put them inside the component. The re-rendering should cause minimal impact on performance.

polymorpher:
I will implement APIs and polish the UI after finishing some other higher priority things first

polymorpher:
Depends on #113 - I will merge that in first on Friday then wrap this up

PR #103, opened 2021-09-01
issue 102: use updated react-qr-reader for restore qr code scan

haolinj:


polymorpher:
Do some field tests and report the result here so we can keep track?

polymorpher:
Selection from multiple camera doesn't work

polymorpher:
fixed

PR #106, opened 2021-09-02
Fix addressinput right click paste on init

haolinj:


haolinj:
@polymorpher this fixes the right-click to paste on address input for initial loading of the select box. But due to the nature of the select input box, as soon as a selection is made, the input value will be covered with the label, which makes it unable to right-click to paste again.

polymorpher:
great, looks good

PR #110, opened 2021-09-02
Revert "issue 102: use updated react-qr-reader for restore qr code scan"

polymorpher:
Reverts polymorpher/one-wallet#103


polymorpher:
Reverting for now. New library does very poorly in dark mode

Issue #102, opened 2021-08-31
Wallet Restore Not Working


roastycoffee:
I'm using 1wallet v0.9.2 in Google Chrome Version 92.0.4515.159 (Official Build) (64-bit)

I created a 1wallet successfully using a Google Authenticator and took note of my wallet address.

I removed the wallet and attempted to do a mock restore of the wallet to test.

I'm asked to input the wallet address, which I successfully supply. Then it asks for access to my computer's camera, which I successfully grant. I can see myself in the camera. I open my Google Authenticator as directed by step 1 and hold the QR code in front of the camera. Nothing happens - it seems like the camera is having a very hard time focusing on the QR code. It seems premature since I'm on step 1 and it doesn't say to "Scan the QR code" until step 3. I don't see how I can advance through the browser wizard?

Regardless, I'm unable to restore my wallet currently.

polymorpher:
I notice this problem from time to time on cameras with lower resolutions and in environments with less contrasting background. The QR code scanner we use is quite dated. I plan to upgrade that component later this week. That should resolve this issue.

polymorpher:
We have tried a few options. Turns out in dark mode, the QR code scanner has some difficulty if your camera is low-res and does not have good auto-focus ability. The most effective solution is in that scenario is holding the QR code really close to the camera, such that the code covers the entire camera region. Ignore the red square. Also, make sure you are selecting only 1 (or 2, if you use double OTP) code. The more codes selected, the more difficult it is for the scanner to pick it up in dark mode. Things would be a lot easier if you switch out of dark mode.

Does it work for?

roastycoffee:
I was using a Logitech c920 which is 1080p in a well let room, my first impression is that it should have worked. Nonetheless, the restore is working now with the new version of the wallet.

PR #112, opened 2021-09-02
Revert "Revert "issue 102: use updated react-qr-reader for restore qr code scan""

polymorpher:
Waiting for https://github.com/BlackBoxVision/react-qr-reader/issues/30 to be resolved.

Should fix https://github.com/polymorpher/one-wallet/issues/102 after the above issue is resolved


polymorpher:
Closing for now - the library requires deeper change at upstream. We may need to write our own upstream library / find a better one for handling QR codes.

PR #117, opened 2021-09-05
Minor responsive fixes

haolinj:


polymorpher:
image

Too much space between title and  "Get Domain"


image

Amounts and "Send" should be aligned right

image

See if you can show two tokens in one row. Alternatively, token panels should be centered and zoom to 100% width if only 1 token per row


image

"Recover" (and Swap, soon to be) tabs are hidden. Can they be displayed in a new line?

image

Can other links be in ... menu as well? This `...` menu seems to only show up when the screen is extra small (iPhone 6/7/8). Perhaps each menu item width can be made smaller, so `...` won't show up even in extra small screen.

 
image

Make title smaller so it is in one line. Or change it to "Create 1wallet"


image

OTP boxes should be bigger and/or more apart (really hard to press).  
"Code for..." can be smaller so the text fits into one line.
"Use two codes..." can also be smaller so it fits into one line.

image

Make upgrade screen not overflow the screen size too much? Right now it is covering the entire card height. Also the text needs to be centered.

image

To and address input should be in one line - make "To" and the address input both smaller so they fit in? The margin near the screen edge is also unncessary.

image

The four elements should be in one line. Make them smaller?

image

The three elements should be in one line

polymorpher:
image

USD amount is not positioned correctly on desktop

polymorpher:
The `...` shows up on iPhone 12 Pro Safari as well, not just extra small screen

IMG_7767

IMG_7766
IMG_7768

polymorpher:
image

width = height?

image

looks weird with the texts so close to border. I think some margin / padding is necessary. I think the left-right margin outside the card (grey area) can be removed

image

same as above. Also applies to other screens (create wallet, restore, recovery, etc.)

image

"Balance" feels redundant. Can we have icon in one row, then <balance> <symbol> in the next row? <symbol> can be in grey color, small font (e.g. `<Hint>` but maybe even smaller than that).  The two rows can be evenly spaced out (e.g. flex) with some space in between. Up to you to fine-tune - let's make it look nice

polymorpher:
image

I think the font and spacing for the tabs can be downsized as well.

image

This bar is too tall - let's shrink it. It is using a lot of screen estate but provides little value

polymorpher:
image

Label texts are cut off. Make them smaller?

polymorpher:
Looks good now. Can we make the mobile view not relying on resize handler and selector? See if pure CSS / React can do the same?

haolinj:
> Looks good now. Can we make the mobile view not relying on resize handler and selector? See if pure CSS / React can do the same?

Do you mean the GridItem height? If we want to make the height the same as the width and width is dynamic fit to the screen, I've tried various ways of using pure CSS, none of them seems to perfectly solve it.

polymorpher:
Let me try some. But first, it is not looking good on desktop anymore. Also there is a weird grey border on the top

image

polymorpher:
Looks good now

PR #120, opened 2021-09-07
Sushiswap - part 1

polymorpher:
Support 100+ pairs / tokens, periodically updated list, and swapping from ONE to tokens and tokens to ONE. Token to token will come in  part 2.

For testing purposes, tokens to ONE is temporarily disabled.


polymorpher:
Looks good. Issues: tokens don’t auto transfer to new wallet after upgrade, because purchasing or swapping tokens on SushiSwap don't cause the tokens to be "tracked" on 1wallet. Only spending those tokens inside 1wallet would track the tokens on the 1wallet smart contract. However, since the new wallet (post-upgrade) can ask the old wallet to transfer the tokens to new wallet, and several functions were designed to do exactly that (partly to deal with this potential pitfall), we can leave this problem to be solved later.

polymorpher:
Another issue is daily limit on the 1wallet contract does not restrict the amount of ONE sent via contract call right now.  I will make another majorVersion upgrade (v11) to fix that, before deploying and make Swap available to everyone in production.

Issue #116, opened 2021-09-04
Red Packet / Gift for offline events


polymorpher:
TODO (deadline Sep 18, 2021):

- [x] Mobile experience optimization (including responsiveness, nav menu presentation, and other adjustments) (@haolinj )
- [x] Fix https://github.com/polymorpher/one-wallet/issues/109 and create mobile OS-specific clickable QR code for auto-deep linking to Google Authenticator
- [x] Gift (Red Packet) UI (under wallet `/show/address]` tab), per documented flow
- [x] UI for claiming funds from Red Packet (a special mobile friendly wallet UI, meant for temporary, shared wallets with limited funds)
- [x] Frontend components for spawning Red Packets
    - [x] Generate a temporary 1wallet
    - [x] Fund the temporary 1wallet and transfer some ERC1155 NFT to it
    - [x] Create sharable link and QR code for other users to claim the red packet (leading to the UI discussed in the step above)
- [ ] Admin tool for minting and managing ERC1155 NFTs for the upcoming event
    - [ ] Standard NFTs for spawning Red Packets
    - [ ] Special NFTs for bonus reward
- [ ] Claimable NFT contracts for the upcoming event
- [ ] (Bonus) Mint NFT from Collectible tab
- [ ] (Bonus) Collectible tab: view 3D NFT  
- [ ] (Non-urgent) UI for claiming bonus event NFT, after the event

polymorpher:
Revised and implemented in https://github.com/polymorpher/one-wallet/releases/tag/v0.12.2 (https://github.com/polymorpher/one-wallet/pull/137)

PR #135, opened 2021-09-15
v12: Extend daily limit to "spending limit"; Allow arbitrary time interval for spending limit; New library "SpendingManager";  Allow multiple contract calls in one operation; Fix various issues

polymorpher:
- Daily limit is superseded by "spending limit", a similar mechanism which allows the spending amount to be reset at arbitrary fixed time interval. This enables use cases such as:
    - Red Packet (see #116), which funds inside the wallet can be withdrawn every 30 seconds or so
    - Parental control by a monthly allowance
    - SaaS subscription
    - and many more
- Spending limit and interval can now be set to any value in expert mode 
- The contract now supports executing multiple operations in one transaction. This means the user only need to type in the authenticator code once to authorize all the operations, instead of typing one code for each operations. Because of that, the user experience in some cases would be significantly improved after we implement corresponding code at the client side. For example, when the user swap some tokens into ONE (or others tokens) using "Swap", they no longer need to make separate authorizations (and wait twice as long) to "Approve SushiSwap to spend your token" then to "Confirm the swap". Now it can be done in one go. 
    - This will be critical for use cases such as Red Packet (#116) since NFT, tokens, and funds need to be transferred simultaneously to the Red Packet 1wallet when the user creates a Red Packet. 
- Many other improvements:
    -  Delete button in address dropdown input box no longer cause the address to be selected and the dropdown menu collapsed (which sometimes results a messy, buggy UI)
    - Address input boxes are now displaying label and domain name of the address correctly after the address is selected
    - Removed erroneous, instant error message (e.g. wallet is outdated) after a new wallet is created or restored.
    - Relayer no longer applies overzealous rate limiting mechanism on errors
    - Fixed an issue on iOS Safari such that the background of the Upgrade screen is not displayed correctly (translucent blur background was not correctly applied)
- Internally:
    - A new library SpendingManager is created. Spending related contract code is moved to the library
    - Nonce related functionalities are moved to CommitManager
    - More abstractions and structures are created  (e.g. CoreSettings), and related code refactored
    - Constructors of 1wallet and its usage in relayer and client are improved
    - Created some library and cache management scripts under relayer


code/lib/api/index.js
```diff
@@ -206,6 +206,21 @@ const api = {
...
+        spendingLimit = r[0]
+        spendingAmount = r[1]
+        lastSpendingInterval = r[2]
+        spendingInterval = r[3]
+      } else {
+        const r = await c.getCurrentSpending()
+        spendingAmount = r[0]
+        lastSpendingInterval = r[1]
+        spendingLimit = dailyLimit
+        spendingInterval = 86400 // default value for pre-v12 wallets i.e. dailyLimit
```
polymorpher, line 221: use ONEConstants

code/contracts/SpendingManager.sol
```diff
@@ -0,0 +1,68 @@
...
+        }
+        return true;
+    }
+
+    function canSpend(SpendingState storage ss, address dest, uint256 amount) external returns (bool){
+        if (!isWithinLimit(ss, amount)) {
+            emit ExceedSpendingLimit(amount, ss.spendingLimit, ss.spentAmount, ss.spendingInterval, dest);
+            return false;
+        }
+        if (address(this).balance < amount) {
```
polymorpher, line 48: should move above line 44

PR #138, opened 2021-09-19
add qr code scanner for address in address input

haolinj:


polymorpher:
Looks very good. I fixed a few things

Issue #155, opened 2021-10-05
can we integrate transak as fiat gateway


lijiang2087:
transak has been a great partner for harmony and they cover many countries and currency as a fiat on-ramp. let's integrate into 1wallet.

--
Link of technical documentation
https://www.notion.so/transak/Integrate-Transak-f43cbb5dc7fa4d068a044ef9589879bf

polymorpher:
Yes, I will test with a non-production key at first

polymorpher:
Is ONE already supported? Didn't see it in the list of selectable cryptocurrencies https://global.transak.com/

polymorpher:
still waiting on transak to fix the issue that crypto gets automatically switched to ETH for US users
![image](https://user-images.githubusercontent.com/913871/140049124-460452a7-42e1-4d5f-8f1e-4394253d25f3.png)

polymorpher:
Deployed using custom made country selector

![telegram-cloud-photo-size-1-4909288215538674161-y](https://user-images.githubusercontent.com/913871/140289269-dbfb9901-1e1e-493c-b302-dae61db9ff67.jpg)

PR #181, opened 2021-11-03
Purchase ONE with fiat (via Transak) 

polymorpher:
Fixes https://github.com/polymorpher/one-wallet/issues/155


polymorpher:
looks good

PR #199, opened 2021-11-26
Fix/persist sync

### LINKED ISSUE #192: Prevent cross-tab wallet state override
Sometimes, newly wallets mysteriously disappear when multiple tabs of 1wallet are opened. This is because each tab is frequently and periodically updating the state of the local database, for example, querying the latest price of ONE/USD from Binance. Right now, everything is bundled into one state space, named `wallet`. As a result, state updates from one tab may override the updates from other tabs, resulting newly created wallets to disappear. Sometimes, this problem extends to completing operations in third-party apps that integrated 1wallet (such as via iframe, in Harmony Safe).

We will add a mechanism to prevent inactive tabs to make any update to the state, so only one tab (in a single thread) may change the state of the database at one time. Over time, we will also separate the singular `wallet` state space into multiple parts, such that balance, price, domain, and address book updates no longer interfere with each other or with wallets themselves.

cylim:
@polymorpher 
- Added `!document.hidden` to stop inactive tab from updating the state and sending message to other tab
- Added crosstab message broadcast and listener to update the redux store whenever new wallet or other state updates

The reason why stopping inactive tab to update the store won't solve the problems,
Tab 1: Created new wallet (persisted to IndexDB)
Tab 2: Redux Store State doesn't have the new wallet, did another action (including price fetcher action), then persisted to IndexDB
Both Tab 1 and Tab 2 lost the new wallet access, because the newer IndexDB data doesn't have the wallet.

closes #192


polymorpher:
Looks good, will test soon

polymorpher:
@hypnagonia told me about https://github.com/pubkey/broadcast-channel, which he recommended and used to fix issues with iframe integrations at https://multisig.harmony.one/

polymorpher:
I looked into this more. Since BroadcastChannel is not supported on Safari (both desktop and mobile), we should use https://github.com/pubkey/broadcast-channel instead

https://developer.mozilla.org/en-US/docs/Web/API/Broadcast_Channel_API#browser_compatibility

polymorpher:
Other than this issue, I think it looks all good. We can insert document.hidden checks to balance queries as well, so to reduce burdens from RPC server (and reduce the chance that the user gets temporary blocked by RPC)

cylim:
All good, the fixes should be straightforward. The exposed API is similar the the browser broadcast channel.

polymorpher:
I think too many rehydrates may cause tabs to crash. In this experiment, I opened two tabs, kept Redux dev tool open for both, and used one tab to browse different wallets. Since opening a wallet fires up a bunch of fetch token (and fetch token successes), the inactive tab crashed. I suggest adding a delay in rehydration and only take the latest state for rehydration if a new update is received cross-tab before the delay expires.

image

polymorpher:
i.e. something similar to lodash debounce (but you might need to implement a custom one)

cylim:
Most actions are just to fetch data every time when user visit certain page, the data doesn't need to be sync with the crosstab feature. Instead I limited to some actions that are more important for the sync to send the message to other browser tabs.

polymorpher:
Okay

polymorpher:
In testing, I often found a series of redundant, mysterious SET_KNOWN_ADDRESSES actions, frequently triggering rehydration of other tabs, but so far it is manageable. This suggests bugs in `AddressInput` component which does these things on initialization @haolinj 

I am also a bit concerned that TRACK_TOKENS is not in allowed actions. I noticed unexpected TRACK_TOKENS actions in my tests but that could be just because I used new wallets on Ganache. Let's test more later and add TRACK_TOKENS and UNTRACK_TOKENS to the list as well.

polymorpher:
Also please eslint next time

PR #200, opened 2021-11-26
Fix/addressbook

### LINKED ISSUE #196: Address book management
Currently, new addresses and their labels are automatically saved to a local state. When addresses are found to correspond to a domain, the domain is used in lieu of a label. Right now, there are ways to delete saved addresses (simply by clicking X when addresses show up), but no way to edit them. Moreover, addresses of deleted wallets are not automatically removed. The result is saved addresses blow up very quickly, and after using the wallets for a while, the users generally see a bunch of irrelevant addresses at first when they click the address input box.  

We should allow addresses to be edited, by adding an edit button, in addition to existing delete button (X). The edit prompt could be a popover, or simply leading the user to a dedicated page where all saved addresses are listed and can be managed. As usual, modal should not be used.
 
References for starting point:
https://github.com/polymorpher/one-wallet/blob/master/code/client/src/components/AddressInput.jsx

cylim:
@polymorpher 
- Added auto remove wallet address in knownAddress
- Added Edit Address Label page for other known address

closes #196

### DIFF
--- a/code/client/src/components/AddressInput.jsx
+++ b/code/client/src/components/AddressInput.jsx
@@ -1,7 +1,8 @@
-import { CloseOutlined, ScanOutlined } from '@ant-design/icons'
+import { CloseOutlined, ScanOutlined, EditOutlined } from '@ant-design/icons'
 import { Select, Button, Tooltip, Row, Col, Spin, Typography, Space } from 'antd'
 import message from '../message'
 import React, { useCallback, useEffect, useRef, useState } from 'react'
+import Paths from '../constants/paths'
 import { useDispatch, useSelector } from 'react-redux'
 import walletActions from '../state/modules/wallet/actions'
 import util, { useWaitExecution, useWindowDimensions, updateQRCodeState } from '../util'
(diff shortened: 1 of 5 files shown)


polymorpher:
thanks, I will try after wrapping up things with a deadline, later today

code/client/src/pages/Contacts/AddressDetail.jsx
```diff
@@ -0,0 +1,122 @@
...
+          <Label ultraWide>
+            <Hint>Address</Hint>
+          </Label>
+          <WalletAddress showLabel address={contact?.address} shorten />
+        </Space>
+        <Space align="baseline" size="large">
+          <Label ultraWide>
+            <Hint>Domain</Hint>
+          </Label>
+          <InputBox
```
polymorpher, line 83: Let's use label since it is immutable

code/client/src/components/AddressInput.jsx
```diff
@@ -302,20 +305,27 @@ const AddressInput = ({ setAddressCallback, currentWallet, addressValue, extraSe
               </Button>
             </Tooltip>
           </Col>
-          <Col span={4}>
+          <Col span={5}>
             {
-            displayDeleteButton
-              ? (
-                <Button
+            displayActionButton
```
polymorpher, line 310: overflowing on mobile

polymorpher:
Overall looks good. I will merge it and do the minor fixes. I will also add you as collaborator in the repository so it would be easier to edit and fix in the future

polymorpher:
Another thing: let's make sure the code is formatted by eslint

Issue #168, opened 2021-10-24
Too Difficult to transfer wallet to another browser from the same computer.  


dastdx1847:
I've tried moving my wallet from a chrome browser to the brave browser.  As previously noted, google authenticator bar code isn't recognized and requires a camera on the device.  Desktop computers may not have cameras attached to them, so an alternative way of inputting the google authenticator code(s) should be made available, particularly for devices that don't have cameras.

polymorpher:
@dastdx1847  give the new import/export feature a try?

PR #202, opened 2021-11-28
add export and import wallet locally #168

cylim:
Related to #168 

Unsure whether should I encrypt the file, the file doesn't contain any sensitive info, and all transaction approval is done with Google Authenticator OTP.


code/client/src/pages/Show/About.jsx
```diff
@@ -144,9 +157,12 @@ const About = ({ address }) => {
           </Col>
         </TallRow>}
       <Row style={{ marginTop: 24 }}>
-        <Popconfirm title='Are you sure？' onConfirm={() => deleteWalletLocally({ wallet, wallets, dispatch, history })}>
-          <Button type='primary' shape='round' danger size='large' icon={<DeleteOutlined />}>Delete locally</Button>
-        </Popconfirm>
+        <Space>
+          <Button type='primary' shape='round' size='large' icon={<ExportOutlined />} onClick={handleExport}>Export locally</Button>
```
polymorpher, line 161: I suggest moving this to the right (justifyContent on outer container)

code/client/src/pages/Show/About.jsx
```diff
@@ -49,6 +50,18 @@ const About = ({ address }) => {
     history.push(Paths.showAddress(address, 'reclaim') + `?from=${backlink}`)
   }
 
+  const handleExport = () => {
+    const label = wallet.name.toLowerCase().split(' ').join('-')
+    const blob = new Blob([JSON.stringify(wallet)])
```
polymorpher, line 55: add mime type for more compatibility?

polymorpher, line 55: layers need to be exported too

code/client/src/pages/List.jsx
```diff
@@ -126,8 +128,54 @@ const List = () => {
...
+      message.error('You can only upload json file')
+    }
+
+    return isJson
+  }
+
   return (
-    <>
+    <Space direction='vertical'>
+      <Upload
```
polymorpher, line 168: I suggest making this its own page, and adding a new top level menu item for it, next to Restore. Or put it in an inconspicuous location at Wallets page

code/client/src/pages/List.jsx
```diff
@@ -126,8 +128,54 @@ const List = () => {
...
+        setFileUploading(false)
+
+        message.info('Imported Successfully')
+      } catch (err) {
+        message.error('File data corrupted')
+      }
+    }
+  }
+
+  const beforeUpload = (file) => {
```
polymorpher, line 157: the layers will be a binary file

code/client/src/pages/List.jsx
```diff
@@ -126,8 +128,54 @@ const List = () => {
       w.address !== ONEConstants.EmptyAddress
   }
 
+  const getBase64 = (data) => new Promise((resolve) => {
+    const reader = new FileReader()
+    reader.addEventListener('load', () => resolve(reader.result))
+    reader.readAsText(data)
+  })
+
+  const handleImport = async (info) => {
```
polymorpher, line 137: probably two files will need to be imported - the layers (partial proofs), and the state (in redux). You can also combine them in one binary blob during export, so import can be done using only one file. To do that, you might need protobuf and define a new .proto.

code/client/src/pages/List.jsx
```diff
@@ -126,8 +128,54 @@ const List = () => {
       w.address !== ONEConstants.EmptyAddress
   }
 
+  const getBase64 = (data) => new Promise((resolve) => {
```
polymorpher, line 131: is it base64?

cylim: It is not. I will change the naming.

code/client/src/pages/List.jsx
cylim: Not very sure about the layers thing... I need to check it out.

code/client/src/pages/List.jsx
cylim: Will move to a new tab on restore page.

polymorpher: I am working on a new Restore page in `#factory` branch. But I agree it makes the most sense to put under Restore page. Let's make sure it is self-contained.

code/client/src/pages/List.jsx
polymorpher: good reference is in `Create.jsx`  - check out what is stored

polymorpher:
Can you make a branch at this repository? So I can push changes too

code/client/src/components/LocalImport.jsx
```diff
@@ -0,0 +1,82 @@
...
+  const handleImport = async info => {
+    if (info.file.status === 'uploading') {
+      setFileUploading(true)
+    }
+
+    if (info.file.status === 'done') {
+      try {
+        const [walletString, ...buffer] = await getDataFromFile(info.file.originFileObj)
+        const wallet = JSON.parse(walletString)
+        const layers = buffer.map(b => new Uint8Array(b.split(',')))
```
polymorpher, line 36: that would be very inefficient - I suggest using protobuf or something here

polymorpher: See https://github.com/protobufjs/protobuf.js/ for encoding and intro

See https://github.com/polymorpher/one-wallet/blob/master/code/client/src/proto/oauth_migration.proto for proto example

See https://github.com/polymorpher/one-wallet/blob/master/code/client/package.json#L10 for compiling

See https://github.com/polymorpher/one-wallet/blob/master/code/client/src/components/OtpTools.jsx#L31 for decoding

cylim: Yep, the current exported files are too huge too, protobuf should able to reduce the size and faster when parsing it. I just upload a WIP version because I need more time on the better solution.

polymorpher: I also wrote a serializer in CLI version without using protobuf: https://github.com/polymorpher/one-wallet/blob/master/code/cli/src/store.js#L41 - that works too but you would need to encode the state part as binary also

cylim:
updated to use protobuf.

code/client/src/components/LocalExport.jsx
```diff
@@ -0,0 +1,49 @@
...
+
+  const handleExport = async () => {
+    try {
+      setLoading(true)
+
+      const layers = await storage.getItem(wallet.root)
+      const filename = `1wallet-${wallet.name.toLowerCase().split(' ').join('-')}.txt`
+      const msg = new LocalExportMessage({ wallet: JSON.stringify(wallet), layers: layers })
+      const buffer = LocalExportMessage.encode(msg).finish()
+      const blob = new Blob([buffer], { type: 'text/plain' })
```
polymorpher, line 29: I think it should be `application/octet-stream` since it is binary data

code/client/src/components/LocalExport.jsx
```diff
@@ -0,0 +1,49 @@
...
+import protobuf from 'protobufjs/light'
+import message from '../message'
+import storage from '../storage'
+import { Button } from 'antd'
+import { ExportOutlined, LoadingOutlined } from '@ant-design/icons'
+import { useState } from 'react'
+
+const Field = protobuf.Field
+
+export function LocalExportMessage (properties) {
```
polymorpher, line 11: I wonder if there are more syntactically cleaner / modern ways of defining it.

code/client/src/components/LocalExport.jsx
```diff
@@ -0,0 +1,49 @@
...
+      const layers = await storage.getItem(wallet.root)
+      const filename = `1wallet-${wallet.name.toLowerCase().split(' ').join('-')}.txt`
+      const msg = new LocalExportMessage({ wallet: JSON.stringify(wallet), layers: layers })
+      const buffer = LocalExportMessage.encode(msg).finish()
+      const blob = new Blob([buffer], { type: 'text/plain' })
+      const element = document.createElement('a')
+
+      element.href = URL.createObjectURL(blob)
+      element.download = filename
+      document.body.appendChild(element)
```
polymorpher, line 34: should remove element after download (click)

code/client/src/components/LocalImport.jsx
```diff
@@ -0,0 +1,74 @@
...
+  const dispatch = useDispatch()
+  const history = useHistory()
+  const wallets = useSelector(state => state.wallet.wallets)
+  const [fileUploading, setFileUploading] = useState(false)
+
+  const getDataFromFile = file =>
+    new Promise(resolve => {
+      const reader = new FileReader()
+      reader.addEventListener('load', () => resolve(reader.result))
+      reader.readAsArrayBuffer(file)
```
polymorpher, line 24: Since this does not throw any error, we should have an error listener as well to reject the promise

code/client/src/components/LocalImport.jsx
```diff
@@ -0,0 +1,74 @@
...
+
+    if (info.file.status === 'done') {
+      try {
+        const data = await getDataFromFile(info.file.originFileObj)
+        const decoded =  LocalExportMessage.decode(new Uint8Array(data))
+        const wallet = JSON.parse(decoded.wallet)
+        const layers = decoded.layers
+
+        if (!util.isValidWallet(wallet)) { throw new Error('Wallet is invalid') }
+        if (util.isWalletExisted(wallets, wallet)) { throw new Error('Wallet is existed') }
```
polymorpher, line 40: just use `wallets[wallet.address]` since keys of wallets are always addresses

code/client/src/components/LocalImport.jsx
```diff
@@ -0,0 +1,74 @@
...
+
+        dispatch(walletActions.updateWallet(wallet))
+        storage.setItem(wallet.root, layers)
+        message.success(`Wallet ${wallet.name} (${wallet.address}) is restored!`)
+        setTimeout(() => history.push(Paths.showAddress(wallet.address)), 1500)
+      } catch (err) {
```
polymorpher, line 46: We should differentiate errors - e.g. failures of reading the file does not mean the wallet is corrupted. It could be device failure (e.g. network/usb drive) or user action (e.g. cancel).

code/client/src/pages/Show/About.jsx
```diff
@@ -144,9 +145,12 @@ const About = ({ address }) => {
           </Col>
         </TallRow>}
       <Row style={{ marginTop: 24 }}>
-        <Popconfirm title='Are you sure？' onConfirm={() => deleteWalletLocally({ wallet, wallets, dispatch, history })}>
-          <Button type='primary' shape='round' danger size='large' icon={<DeleteOutlined />}>Delete locally</Button>
-        </Popconfirm>
+        <Space>
```
polymorpher, line 148: Let's have the items spread out (flex space-between), and delete action on the left, so it is consistent with overall style. (negative action on the left, positive on the right)

code/client/src/components/LocalImport.jsx
```diff
@@ -0,0 +1,74 @@
...
+
+  return (
+    <Upload
+      name='walletjson'
+      showUploadList={false}
+      customRequest={({ onSuccess }) => { onSuccess('ok') }}
+      accept={'text/plain'}
+      onChange={handleImport}
+    >
+      <Button
```
polymorpher, line 62: This should be placed below the main content, and with some explanations of what it is and what is to be expected. But I can work on those later, since I am making a new Restore page anyway in #factory

code/client/src/components/LocalExport.jsx
cylim: It seems like the file extension is affecting the mime-type, especially for Safari. I tried to use .pb or other extension, it ended up the mime type is empty.

code/client/src/components/LocalExport.jsx
cylim: I will look into it. It has multiple ways to define the message.

cylim: Although there are multiple ways to declare the proto type, none of it is cleaner than the current option. The only cleaner/modern option is using decorator with TypeScript.

https://github.com/protobufjs/protobuf.js#examples

polymorpher:
Okay, perhaps later when it's more complex, we can define its own .proto file and compile it to js

polymorpher:
I wrote https://github.com/polymorpher/one-wallet/blob/factory/code/client/src/proto/wallet.proto for another use case of export/import. Works great. I will merge these two implementations later. Also I think some explanations and messages should be added in export to caution users keeping the exported state file secure. If they are leaked, it weakens  the security by quite a lot. Even though the attacker can't take control of the wallet without the authenticator code, he may crack it by spending enough resources. See https://github.com/polymorpher/one-wallet/wiki/Client-Security

cylim:
layers are repeated, it should be like
```
repeated bytes layers = 3
```

polymorpher:
image

polymorpher:
image

polymorpher:
image

Issue #204, opened 2021-12-05
Convert 1USDC to ONE gives you wrong value


dietbald:
When trying to swap 1 1USDC to ONE , it is also impossible to complete the swap 

![image](https://user-images.githubusercontent.com/1293151/144751137-e3ecd379-3a3b-404d-9fad-96e527a27ae5.png)

polymorpher:
Was the amount automatically filled? It does not appear so. Also you do not have any 1USDC balance to swap from

dietbald:
I do have balance in the wallet for it. I thought I actually transfered 1 1USDT and 1 1USDC to the wallet, but I might have been wrong. I also dont seem to see the transaction in the block explorer in the HRC20 tab

![image](https://user-images.githubusercontent.com/1293151/144751254-0ebf0473-e070-49d7-83d6-32307e4c23cd.png)

polymorpher:
I will look into this. I think it might be the balance in Swap page is out of sync with the balance on main page. There may be an error on the number of decimal points that are supposed to be used. And the large number is caused by either that or a divide by zero error.

polymorpher:
@cylim  if you have time - feel free to take a look as well

dietbald:
1usdc, 1usdt and AAVE seem to have similar issue 

![image](https://user-images.githubusercontent.com/1293151/144752368-26301fcc-da0b-4c7c-88db-1b9f31b95fba.png)
![image](https://user-images.githubusercontent.com/1293151/144752380-c21a9301-058d-4d18-8e25-1465112fc231.png)

cylim:
@dietbald when you select the usdc, there are two USDC, select the one with squared icon instead of the rounded icon.

<img width="186" alt="Screen Shot 2021-12-06 at 16 05 17" src="https://user-images.githubusercontent.com/5622951/144809900-851bdd41-c875-4515-bc95-5bd5ab141832.png">

cylim:
@polymorpher  After removing the default tokens, the AAVE pricing still incorrect, probably because of liquidity issues in Sushi Swap. I need to spend more time to look into it.

And we might need to update icons for some tokens.

dietbald:
@cylim  I get the error for both of the Tokens, see in the picture, also the balance doesnt seem to come through as well 

![image](https://user-images.githubusercontent.com/1293151/144842781-8e404ae7-6318-465d-ad06-f5dd7444f1f4.png)
[...]

PR #205, opened 2021-12-06
add decimals for default known erc20 tokens

cylim:
Removed default  tracked erc20 token

Reasons,
1. User doesn't have any tokens with those, if they do have some, it will be part of `erc20Tracked`
2. Those known erc20 token doesn't have decimals - https://github.com/polymorpher/one-wallet/blob/e9f8edc4c123fb104c159c7544a40806ee09aaed/code/client/src/components/TokenAssets.jsx#L16-L22

related #204

### DIFF
--- a/code/client/src/pages/Show/Swap.jsx
+++ b/code/client/src/pages/Show/Swap.jsx
@@ -208,7 +208,8 @@ const Swap = ({ address }) => {
       return
     }
     const erc20Tracked = (wallet.trackedTokens || []).filter(e => e.tokenType === ONEConstants.TokenType.ERC20)
-    const trackedTokens = [harmonyToken, ...withKeys(DefaultTrackedERC20(network)), ...(erc20Tracked || [])]
+    const trackedTokens = [harmonyToken, ...(erc20Tracked || [])]
+    console.log(trackedTokens)
     trackedTokens.forEach(tt => {
       // align formats
       tt.address = tt.address || tt.contractAddress
--- a/code/client/src/pages/Show/Swap.jsx
+++ b/code/client/src/pages/Show/Swap.jsx
@@ -208,8 +208,7 @@ const Swap = ({ address }) => {
       return
     }
     const erc20Tracked = (wallet.trackedTokens || []).filter(e => e.tokenType === ONEConstants.TokenType.ERC20)
-    const trackedTokens = [harmonyToken, ...(erc20Tracked || [])]
-    console.log(trackedTokens)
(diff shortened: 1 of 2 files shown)


cylim:
I changed my mind, adding decimal is better than removing the default token, due to the token might not tracked. I am looking into how to remove duplicated token in the list.

polymorpher:
The decimals are actually retrieved from https://github.com/polymorpher/one-wallet/blob/master/code/client/src/pages/Show/Swap.jsx#L227 , similarly in https://github.com/polymorpher/one-wallet/blob/master/code/client/src/components/ERC20Grid.jsx#L40 . But it is not a bad idea to hard code them in the list so the decimals are put in use immediately.

Also the list of ERC20 tokens are taken from top ones in https://explorer.harmony.one/hrc20. The purpose to include them is to promote and showcase those

code/client/src/pages/Show/Swap.jsx
```diff
@@ -208,7 +208,7 @@ const Swap = ({ address }) => {
       return
     }
     const erc20Tracked = (wallet.trackedTokens || []).filter(e => e.tokenType === ONEConstants.TokenType.ERC20)
-    const trackedTokens = [harmonyToken, ...withKeys(DefaultTrackedERC20(network)), ...(erc20Tracked || [])]
+    const trackedTokens = [harmonyToken, ...(erc20Tracked || []), ...withKeys(DefaultTrackedERC20(network))]
```
polymorpher, line 211: I think default ones should go first, so the tracked ones can override them if there is any update

code/client/src/pages/Show/Swap.jsx
```diff
@@ -223,8 +223,10 @@ const Swap = ({ address }) => {
         dispatch(walletActions.fetchTokenBalance({ address, tokenType, tokenId, contractAddress, key }))
       }
     })
+    const filteredTrackedTokens = [...new Map(trackedTokens.map(v => [v.address, v])).values()]
```
polymorpher, line 226: Could use lodash uniqBy

code/client/src/pages/Show/Swap.jsx
cylim: The issue is the tracked token might not have icons from the sushiswap cloudinary, e.g. 1aave

polymorpher: Let's move icons to a GCP storage bucket later

code/client/src/pages/Show/Swap.jsx
cylim: Yep, the project has lodash, I will change it later. :)

polymorpher: I changed it in master

Issue #198, opened 2021-11-25
Refactoring local state


polymorpher:
Right now, all local state (aside from partial proofs) are stored under a single `wallet` object. See https://github.com/polymorpher/one-wallet/blob/master/code/client/src/state/modules/wallet/reducers.js

This causes several issues.  For example, https://github.com/polymorpher/one-wallet/issues/192, and that it is very easy to mess up the entire state in a simple, async local update, such as changing a simple flag (e.g. whether wallet QR code was saved) in one single wallet. Global states (such as address book) and states of each individual wallet are stored together under this object.

This should not be the case. We should refactor the whole redux flow and keep each storage object as minimal as possible.

PR #219, opened 2021-12-18
State refactor #3

taoalpha:
https://github.com/polymorpher/one-wallet/issues/198


polymorpher:
change base branch?

polymorpher:
Looks good, but it is not backward compatible. Users would see their wallets disappear and panic. Auto-migration is needed

taoalpha:
> Looks good, but it is not backward compatible. Users would see their wallets disappear and panic. Auto-migration is needed

DONE

polymorpher:
On a second thought, let's merge this first, since there will be a lot more changes later

polymorpher:
The left-over variables in `wallet` are selected and assumed as wallets when `useSelector(state => state.wallet)` are used. Some pages fail to render because of this, for example, `List.jsx`

polymorpher:
Fixed. Seems fine now.

Issue #220, opened 2021-12-18
Allow restore from laptop by uploading authenticator qr code


taoalpha:
This will make testing restore locally much easier

PR #221, opened 2021-12-18
Allow restore by uploading QR image

taoalpha:
https://github.com/polymorpher/one-wallet/issues/220


code/client/src/components/QrCodeScanner.jsx
```diff
@@ -144,7 +148,18 @@ const QrCodeScanner = ({ onScan, shouldInit, style }) => {
               </Row>
             </>
             )
-          : <></>
+          : (
+            <Upload
```
polymorpher, line 152: deduplicate?

Issue #176, opened 2021-10-28
Informative error messages from emitted logs


polymorpher:
Many users are confused when 1wallet prompts their transaction is "done", but in reality the transaction actually failed. For example, when the user attempts to send ONE beyond what their daily limits allow them to do, they would assume the transaction is successful, but it is in fact not so. Many users are unaware of the existence of such limit, which by default is 1000 ONE per day. The transaction https://explorer.harmony.one/tx/0x224d03f05e109e4ef71ebf6ae4d68a4dd0822f775584154cf05217f9be06c36b?activeTab=2 is an example. The error is only revealed through the log, which shows topic `0x55e3e9bfe128ba9adab893485d7dcbbbce711ba4ae2ab5154181467093f1d423`, that maps to event `ExceedSpendingLimit(uint256,uint256,uint256,uint256,address)`. See also https://github.com/polymorpher/one-wallet/blob/master/code/scripts/events-map.txt. 

The confusion comes from the fact that "done" only means the transaction is completed. It does not mean the operation necessarily succeeded. For security reasons (to prevent OTP re-use), 1wallet does not revert a transaction when authentication is successful but the operation is not. This mechanism is different from other wallets and common practice in smart contract, which simply reverts transactions when the intended operation fails. 

Accordingly, it is necessary to decode transaction receipts (which are returned by relayer after a transaction is completed) and display its content in human readable form to the user (within notification boxes). In addition to simple send, other operations may also result in error (such as recovery, upgrade, and others). Errors should be parsed, displayed promptly and intelligently, instead of expecting the user to decipher transaction logs in explorer.

cylim:
@polymorpher 
Had spent some time understanding the current system, I am not sure what is the approaches to this issues.

Tried to check the receipts from the transaction, the event logs are not in the `tx.receipt` yet. Probably have to do another call to `method: hmyv2_getTransactionReceipt`, unsure whether if calling it right after the transaction, the log might be empty?

trx: [https://explorer.harmony.one/tx/0x58cbca4ac3628c5ba83c41d0a4416a630f977218a1a245fbf9c8a0fad058060d?activeTab=2](https://explorer.harmony.one/tx/0x58cbca4ac3628c5ba83c41d0a4416a630f977218a1a245fbf9c8a0fad058060d?activeTab=2)

[...]

PR #222, opened 2021-12-19
Event logs

cylim:
#176 
@polymorpher I am not sure whether this is the request approach to the issues. Let me know if there is better approaches.


code/scripts/eventhash.js
```diff
@@ -1,33 +1,33 @@
...
   const hashMap = {}
+  const hashMapJson = {}
   for (let line of lines) {
     const m = line.match(PATTERN)
     if (!m) {
       continue
     }
     const method = m[1]
     const args = m[2]
-    if (!args) {
```
polymorpher, line 17: events with empty argument would have `args.split(', ')` to be `['']` and compute the incorrect hash. If this block is removed, `paramTypes` needs to consider this case later

polymorpher:
Thanks. I think overall this mechanism is fine.

code/scripts/eventhash.js
cylim: About the issue you mentioned. It doesn't end up with two different hash, if it does, you would have `two records with different hash and same event name` in the output(event-map.txt) file, because it is `if statement, not if else statement` the if statement **added into hashMap** and then **continue to run the method to find hash with params and add into the hashMap**.  


''.split(', ') = ['']
[''].join(',') = ''

polymorpher: Okay, let me verify and check this later. I can't remember the exact reason I did it like that before, but I recall there are some.

polymorpher: yes

Issue #206, opened 2021-12-08
Batch Redux state updates, lazy load wallet components


polymorpher:
## Batch Redux state updates

When a wallet is opened by the user, a large number of state updates are dispatched. For example, the balance update for each tracked (and default) ERC20 tokens. This causes mildly visible slowdown of the UI, and should be fixed as soon as possible. Other operations similar to this should also be batched: wallet balance updates, for instance.

## Lazy load wallet components

The wallet component (https://github.com/polymorpher/one-wallet/blob/master/code/client/src/pages/Show.jsx) loads a large number of subcomponents, as soon as the user enters the wallet. Many of these components are not going to be used, and should be lazily loaded.

PR #224, opened 2021-12-20
Batch state updates #1

taoalpha:
https://github.com/polymorpher/one-wallet/issues/206


polymorpher:
looks good, have you had a chance to look at https://github.com/polymorpher/one-wallet/pull/199#issuecomment-991871506 and see if the fundamental issues with SET_KNOWN_ADDRESS can be fixed? It shouldn't be called every time the page loads.

taoalpha:
will look into that later

polymorpher:
Redux batch seems to be designed to minimize render updates to React. (https://react-redux.js.org/api/batch). The changes you made fixed that and should speed up things a lot for rendering. Another concern discussed in the comment linked above is related to cross-tab propagation issues - whenever a state update is triggered, a cross-tab message is sent to other tabs and other tabs replaces the entire Redux store's state with the content of the message. If the state updates are too frequent, other tabs are overwhelmed with updates and may crash. Batching does not seem to reduce the number of state updates.

code/client/src/components/AddressInput.jsx
```diff
@@ -163,39 +163,50 @@ const AddressInput = ({ setAddressCallback, currentWallet, addressValue, extraSe
...
-
-      unlabelledWalletAddress.forEach((w) => {
-        dispatch(globalActions.setKnownAddress({
-          ...knownAddresses[w],
-          label: w.name,
-        }))
-      })
-
-      await Promise.all(knownAddressesWithoutDomain.map(async (knownAddress) => {
+      const domainWalletAddresses = await Promise.all(knownAddressesWithoutDomain.map(async (knownAddress) => {
```
polymorpher, line 166: There could be a long delay in resolving this. We shouldn't wait for this to finish before updating the rest

polymorpher:
Other parts look good. We can merge this first (after the commented issue is fixed) and deal with cross-tab issue in separate PR

code/client/src/components/AddressInput.jsx
taoalpha: Done

PR #225, opened 2021-12-20
Lazyload sections

taoalpha:
https://github.com/polymorpher/one-wallet/issues/206


polymorpher:
Have you tested? I recall there was some reason that these components are controlled by a `show` parameter, instead of just doing the way you did. Perhaps it was just for transition animation, but I am not sure. It was too long ago. (June 2021)

taoalpha:
> Have you tested? I recall there was some reason that these components are controlled by a `show` parameter, instead of just doing the way you did. Perhaps it was just for transition animation, but I am not sure. It was too long ago. (June 2021)

ya I tested with `domain`, the dialog shows correctly. And from the code, the `show` param from AnimatedSection is default to true: https://github.com/polymorpher/one-wallet/blob/96e9f772658206f470f3f8081bfb1db54170958f/code/client/src/components/AnimatedSection.jsx#L52

polymorpher:
Okay. Seems fine when I tested

Issue #119, opened 2021-09-06
Remove Truffle and its artifacts


polymorpher:
The current client size bundle is 8.8MB (+0.5MB for workers, before compression). It improved a lot (from ~15MB before), but is still quite large. This is because we used Truffle for loading contracts and making calls to the blockchain. Since the calls are read-only, and we do not need any features provided by Truffle that are not provided by ethers or web3, we can cut off Truffle and artifacts generated by Truffle from client side dependencies. This should reduce client size by 50-75%, which would significantly improve loading speed.

Replacement candidates are 
https://bundlephobia.com/package/ethers@5.4.6 (~450KB) 
https://bundlephobia.com/package/web3@1.5.2 (1.1MB)

I am more inclined to use ethers, but web3 would be less disruptive since Truffle is based on web3. We can transition to ethers later.

polymorpher:
Related: https://github.com/polymorpher/one-wallet/issues/2 (analysis is quite outdated)

PR #228, opened 2021-12-24
reduce the bundle size

taoalpha:
modular lodash and antd + ant-design to reduce the bundle size

ideally this should be done by babel-plugin-import (partially), but it seems not work well with react 17, so update manually instead

https://github.com/polymorpher/one-wallet/issues/119


taoalpha:
have to change the call signature for some of the methods due to the difference between the main and lp: https://github.com/lodash/lodash/wiki/FP-Guide#fixed-arity

polymorpher:
It's good but impact seems to be very small. Let's focus on high impact stuff first

taoalpha:
> It's good but impact seems to be very small. Let's focus on high impact stuff first

ya just some low hanging fruit

taoalpha:
also from the bundle size, the ant + highlight account for 30-40% of the total size, the truffle itself seems pretty slim 

image

polymorpher:
This can be almost entirely removed by taking out Truffle, in addition to itself

image
image

image

ENS related stuff seems to be taking an unreasonable amount of space (idna map, @ensdomain, unorm, etc.). We should take them out too and use custom implementation.

Similarly, perhaps we don't need html2canvas (could use html-to-image), react-player (just use HTML5 video), moment, and text-encoding.

Not sure why app.less takes so much space, is the image size not proportional to actual size?

taoalpha:
ya I am looking into the truffle stuff, meanwhile, simply just by replacing all antd with antd/es can benefit us ~1Mb more, down to 
```
Assets: 
  main.js (9.05 MiB)
  ONEWalletWorker.js (675 KiB)
```

I think the app.less is also bloated by the antd styles...

image

taoalpha:
after modular ant-design/icons:

```
Entrypoints:
  main (8.57 MiB)
      main.js
  ONEWalletWorker (675 KiB)
      ONEWalletWorker.js
```

PR #245, opened 2022-01-23
Some minor ux fix

taoalpha:


code/client/src/components/SiderMenu.jsx
```diff
@@ -73,6 +73,10 @@ const SiderMenu = ({ ...args }) => {
     }
   }, [])
 
+  useEffect(() => {
```
polymorpher, line 76: is this needed? I think the selector automatically updates the state

PR #292, opened 2022-04-16
Refactor Create

taoalpha:
This will make it easier to refactor the code for v2ui

### DIFF
--- a/code/client/src/pages/Create.jsx
+++ b/code/client/src/pages/Create.jsx
@@ -79,11 +79,8 @@ const Create = ({ expertMode, showRecovery }) => {
   const [spendingInterval, setSpendingInterval] = useState(WalletConstants.defaultSpendingInterval) // seconds, number
 
   const [worker, setWorker] = useState()
-  const [root, setRoot] = useState()
-  const [hseed, setHseed] = useState()
-  const [layers, setLayers] = useState()
-  const [innerTrees, setInnerTrees] = useState()
-  const [slotSize, setSlotSize] = useState(1)
+
+  const [rhlis, setRhlis] = useState({})
   const [progress, setProgress] = useState(0)
   const [progressStage, setProgressStage] = useState(0)
   const [address, setAddress] = useState() // '0x12345678901234567890'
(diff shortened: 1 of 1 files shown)


code/client/src/pages/Create.jsx
```diff
@@ -45,108 +44,84 @@ const getGoogleAuthenticatorAppLink = (os) => {
...
+    seed2: generateOtpSeed(),
+    duration: WalletConstants.defaultDuration,
+    doubleOtp: false
+  }).current
 
   const [effectiveTime, setEffectiveTime] = useState()
+  const [otpReady, setOtpReady] = useState(false)
 
-  const generateNewOtpName = () => ONENames.genName(Object.keys(wallets).map(k => wallets[k].name))
-  const [name, setName] = useState(generateNewOtpName())
```
polymorpher, line 67: Let's keep name. We should allow users to pick their own name and dynamically change the QR code

taoalpha: ya but it's not used yet, and we can add it back whenever we need, like in the v2ui. But for now, I want to remove those unused stuff so we can keep it simple for the upgrade

polymorpher:
It looks good, but I think `address` and `walletInfo.doubleOtp` should also be made state variables instead of native variables. Rendering components and functions defined on a per-render basis depend on those. In the case of `address`, being a native variable is harmless because we know by the time `walletInfo.address` is rendered and accessed (when DoneSection is shown), it was already set and won't be changed. But it can be problematic for `doubleOtp`. When its value is changed, components and functions won't be re-rendered and re-computed automatically

polymorpher:
also lots of this error
image

polymorpher:
With two codes, it also goes straight to deployment, rather than waiting for user to confirm (and change spending limit if needed)

polymorpher (approved):
I made some changes and should have fixed the issues. Do some tests? Especially cases of expert-mode enabled, and scanning is done quickly before worker is done

polymorpher:
fixed it

PR #294, opened 2022-04-17
feat(v2ui): setup Restore for v2ui

BrayOOi:
Add RestoreV2 component to house new design requirements for Restore page


code/client/src/pages/Restore.jsx
```diff
@@ -128,4 +129,134 @@ const Restore = () => {
...
+          case Sections.SyncRecoveryFile:
+            return (
+              <AnimatedSection>
+                <SyncRecoveryFile
+                  onSynced={onSynced}
+                  onCancel={() => setSection(Sections.Choose)}
+                />
+              </AnimatedSection>
+            )
+          case Sections.SetupNewCode:
```
polymorpher, line 206: Need this component to render for `section === Sections.SetupNewCode || section === Sections.RecoveryCode` because it computes something in the background (then triggers a callback function to sync the result) while the user is on the next page (Sections.RecoverCode)

polymorpher:
@taoalpha what do you think?

code/client/src/pages/Restore.jsx
taoalpha: the switch case can be handy sometimes, but may not be the best choice for this, personally I would avoid too many indentation levels, and sometimes does not work that great like the case pointed above.

I would say extract the sections below into separate components and then just use either `if`s or merge them into one return or do both.

code/client/src/pages/Restore.jsx
```diff
@@ -128,4 +129,134 @@ const Restore = () => {
...
+  const [innerTrees, setInnerTrees] = useState()
+  const [innerCores, setInnerCores] = useState()
+  const [expert, setExpert] = useState()
+  const [name, setName] = useState()
+  const [newLocalParams, setNewLocalParams] = useState()
+  const [address, setAddress] = useState()
+
+  const onSynced = async ({ name, address: retrievalAddress, innerTrees, expert }) => {
+    try {
+      const { wallet } = await retrieveWalletInfoFromAddress(retrievalAddress)
```
taoalpha, line 150: if all the set ups are the same, like the only change is the dom structure, I would do the upgrade support from the Restore component. Of course you will need to extract the components from Restore to multiple components first (like per section), but that should make it more readable and modular as well ;) we should try to avoid very large components in general, they tend to become really hard to maintain pretty soon, especially when no tests ;)

taoalpha:
> @taoalpha what do you think?

added a few comments

taoalpha:
I just merged the basic setup pr and deleted the newui branch, it automatically closed this one since it's targeting that branch... you can reopen this and point to master

polymorpher:
Is this being worked on?

polymorpher:
Let's reopen this when it is ready for review

PR #306, opened 2022-04-25
Autofill OTP setup, backend, account signup, login

polymorpher:


polymorpher:
To merge after https://github.com/polymorpher/one-wallet/issues/309 fix is done and deployed

polymorpher:
Note: after this is merged, client will attempt to relayer and backend HTTPS ports by default instead of HTTP based. Open your browser and access `/health` routes of the corresponding server first, so that the browser won't have any issue sending HTTPS ajax requests to them later. This is mandatory for Safari

polymorpher:
Self-reviewed. Looks good. Need to add sign-up form in Restore as well.

PR #310, opened 2022-04-26
v2 routes + components

taoalpha:


polymorpher:
Changes look fine, pending tests. Not sure what some of the v2 grids are doing and whether routes are implemented to cover all cases, but will find out more during tests

polymorpher:
I think wallet selector should still be shown in wallet-creation screen, if there are existing wallets. The wallet selector's create button is also only sensitive to the text itself instead of the entire row. The "Assets" page is currently somewhat hard to interpret.

Going to merge for now. Let's fix later

polymorpher (approved):

PR #314, opened 2022-04-30
Minor fix on tracked tokens

taoalpha:
change how the currentTrackedTokens is updated, this will fix the issue that the tokens added from UI is lost after refresh


code/client/src/components/ERC20Grid.jsx
```diff
@@ -205,12 +199,11 @@ export const ERC20Grid = ({ address }) => {
...
         }
       }))
-      // console.log('tts merged', tts)
       if (cancelled) {
         return
       }
-      // tts = uniqBy(tts, t => t.key)
-      setCurrentTrackedTokens(tts)
+      // Merge to existing list.
+      setCurrentTrackedTokens(ptts => uniqBy(t => t.key, [...ptts, ...tts]))
```
polymorpher, line 206: should still filter out `untrackedTokenKeys` and merge with `defaultTrackedTokens`

taoalpha: done

polymorpher (approved):

PR #317, opened 2022-05-21
fix: fallback to wallet on lost oldInfos

SystemDisc:
@polymorpher This should do it! Where do I see the bounty?

### DIFF
--- a/code/lib/api/flow.js
+++ b/code/lib/api/flow.js
@@ -80,7 +80,7 @@ const EOTPDerivation = {
           if (idKeyIndex === identificationKeys.length - 1) {
             layers = await storage.getItem(root)
           } else {
-            const info = oldInfos[idKeyIndex]
+            const info = oldInfos[idKeyIndex] ? oldInfos[idKeyIndex] : wallet
             layers = await storage.getItem(info.root)
             effectiveTime = info.effectiveTime
             root = info.root


polymorpher:
Thanks, but why should it fallback to wallet itself in this case? It must have found an identification key in this case that does not match root and matches with old cores (and their roots) in this case

SystemDisc:
@polymorpher Yes, perhaps. Maybe I can share my screen with you and show you the issue I'm encountering. I think that may help us get to the crux of the problem.

polymorpher:
Yes, thanks, that would be helpful. You can also share some screen recordings. What's the wallet address?

polymorpher:
I sent 1000 ONE as my appreciation for debugging and reporting the issue that the relayer went down, and this issue. It's not a lot given the current market condition, but I hope it is helpful. At this time, there is no bounty program with this project. I am paying out of pocket for this, though perhaps Harmony may be able to reimburse some later

The transaction is
https://explorer.harmony.one/tx/0xdacb1597f41837d70a7190acfab0ad9c15cc168e8741edc87a2f51c78b02add8

SystemDisc:
@polymorpher I'll email you

urica12 (approved):
code/lib/api/flow.js

urica12 (approved):
approve

urica12 (changes requested):
looks good i approve this changes

urica12 (approved):
approve this chang

PR #362, opened 2024-01-06
Recovery file backup system, client, and UI implementations; Add Wallet Connect to wallet page; Reworked some wallet UI, some restore text and UI; Renaming some tabs; Hiding deprecated integrations

polymorpher:


polymorpher:
Self reviewed, looks good

